PR #4, opened 2023-04-19
Added HVA.py

simonleandergrimm:
I now added HVA.py, a script calculating the prevalence of Hepatitis A virus for Seattle in 2017 and 2018, and for the US in 2018. 

Let me know what you think, I'd be particularly keen to make the code more efficient (for instance, I'm referring to the `prevalence_est` dictionary (prevalence estimators) 7 times in the final calculation).


pathogens/hva.py
```diff
@@ -0,0 +1,100 @@
...
+    taxid=208726,
+)
+
+prevalence_est = {  # prevalence estimators
+    "us_incidence_absolute_2018": PrevalenceEstimator(
+        value=12474,
+        unit="cases",
+        value_type="incidence_absolute",
+        country="United States",
+        source="https://www.cdc.gov/hepatitis/statistics/2018surveillance/index.htm",
```
jeffkaufman, line 23: Is there a more specific URL?  I don't see 12474 on that page

pathogens/hva.py
```diff
@@ -0,0 +1,100 @@
...
+        start_date="2018-01-01",
+        end_date="2018-12-31",
+    ),
+    "us_population_2018": PrevalenceEstimator(
+        value=326.8 * 1e6,
+        unit="people",
+        value_type="population",
+        country="United States",
+        source="https://www.google.com/search?q=us+population+2018&oq=us+population+2018&aqs=chrome.0.69i59j0i512j0i22i30l6j0i390i650l2.5937j1j7&sourceid=chrome&ie=UTF-8",
```
jeffkaufman, line 32: The `&oq=us+population+2018&aqs=chrome.0.69i59j0i512j0i22i30l6j0i390i650l2.5937j1j7&sourceid=chrome&ie=UTF-8` isn't doing anything

pathogens/hva.py
```diff
@@ -0,0 +1,100 @@
...
+    "us_incidence_absolute_2018": PrevalenceEstimator(
+        value=12474,
+        unit="cases",
+        value_type="incidence_absolute",
+        country="United States",
+        source="https://www.cdc.gov/hepatitis/statistics/2018surveillance/index.htm",
+        start_date="2018-01-01",
+        end_date="2018-12-31",
+    ),
+    "us_population_2018": PrevalenceEstimator(
```
jeffkaufman, line 27: Doesn't have to be fixed now, but at some point we'll want a way to pull commonly referenced numbers like the US population out into something shared

pathogens/hva.py
jeffkaufman, line 32: I guess citing "google" is ok for numbers this basic, though I'd feel slightly better about citing the actual primary source, which in this case is the census

pathogens/hva.py
```diff
@@ -0,0 +1,100 @@
...
+        value=326.8 * 1e6,
+        unit="people",
+        value_type="population",
+        country="United States",
+        source="https://www.google.com/search?q=us+population+2018&oq=us+population+2018&aqs=chrome.0.69i59j0i512j0i22i30l6j0i390i650l2.5937j1j7&sourceid=chrome&ie=UTF-8",
+        start_date="2018-01-01",
+        end_date="2018-12-31",
+    ),
+    "hva_shedding_duration": PrevalenceEstimator(
+        value=14,
```
jeffkaufman, line 37: The source has `1-3 weeks` -- I think we should include that range here, and have the code that works with the variable take the midpoint if it wants

pathogens/hva.py
```diff
@@ -0,0 +1,100 @@
...
+    "king_county_incidence_rate_2017": PrevalenceEstimator(
+        value=0.5,
+        unit="cases_per_100k",
+        value_type="incidence_rate",
+        country="United States",
+        state="Washington",
+        county="King",
+        start_date="2017-01-01",
+        end_date="2017-12-31",
+        source="https://doh.wa.gov/sites/default/files/2023-01/420-004-CDAnnualReport2021.pdf?uid=642c448518316#page=28",
```
jeffkaufman, line 52: It looks like you're interpreting this as "infections / population", but isn't it actually "confirmed cases / population"?  In which case don't we need to adjust for underreporting?

pathogens/hva.py
```diff
@@ -0,0 +1,100 @@
...
+    ),
+    "hva_shedding_duration": PrevalenceEstimator(
+        value=14,
+        unit="days",
+        value_type="shedding_duration",
+        country="United States",
+        source="https://www.cdc.gov/vaccines/pubs/pinkbook/hepa.html#:~:text=Viral%20shedding%20persists%20for%201%20to%203%20weeks.",
+    ),
+    "king_county_incidence_rate_2017": PrevalenceEstimator(
+        value=0.5,
```
jeffkaufman, line 44: We should also copy over the absolute number of cases, since that's useful for estimating the error

pathogens/hva.py
```diff
@@ -0,0 +1,100 @@
...
+
+    incidence_rate = (absolute_incidence / population) * 100000
+
+    prevalence = incidence_rate * (shedding_duration / 365)
+
+    return prevalence
+
+
+prevalence_vars = {
+    "us_prevalence_2018": calculate_prevalence_from_absolute_incidence(
```
jeffkaufman, line 87: Instead of doing `var.value` everywhere, pass the `PrevalenceEstimator` to the function and have it do the work.

Additionally, `calculate_prevalence_from_incidence_rate` can verify that the `value_type` it's been given is correct.

jeffkaufman: Possibly we should use the type system more and every distinct `value_type` should be a different class?  Then we can use `mypy` for static typechecking?

pathogens/hva.py
```diff
@@ -0,0 +1,100 @@
...
+
+def calculate_prevalence_from_absolute_incidence(
+    absolute_incidence: int, population: int, shedding_duration: int
+) -> float:
+
+    incidence_rate = (absolute_incidence / population) * 100000
+
+    prevalence = incidence_rate * (shedding_duration / 365)
+
+    return prevalence
```
jeffkaufman, line 83: These two functions should be somewhere more general where they can be shared, though you don't have to fix this in this PR.

pathogens/hva.py
```diff
@@ -0,0 +1,100 @@
...
+) -> float:
+    prevalence = incidence_rate * (shedding_duration / 365)
+    return prevalence
+
+
+def calculate_prevalence_from_absolute_incidence(
+    absolute_incidence: int, population: int, shedding_duration: int
+) -> float:
+
+    incidence_rate = (absolute_incidence / population) * 100000
```
jeffkaufman, line 79: Not good to have "10000" here without anything connecting it to `unit="cases_per_100k"`.

jeffkaufman (changes requested):

jeffkaufman:
Separately: you don't need to put "ready for review by X" in the title.  You can use the Reviewers system to specifically request a review from someone, and if you don't want anyone looking at your PR you can [convert it to a draft](https://github.blog/changelog/2020-04-08-convert-pull-request-to-draft/) (see the upper right hand corner of this page for what that link looks like)

pathogens/hva.py
```diff
@@ -0,0 +1,100 @@
...
+sys.path.append("..")
+
+from p2ra_prevalences.pathogen_properties import *
+import math
+
+background = """Hepatitis A is a vaccine-preventable liver-infection caused by the hepatitis A virus. In the US,
+it's mostly spread by individual contact. There is little seasonal variance in Hepatitis A incidence (https://www.cdc.gov/vaccines/pubs/pinkbook/hepa.html). Viral shedding persists for 1 to 3 weeks."""
+
+pathogen_chars = PathogenChars(
+    na_type=NAType.RNA.value,
```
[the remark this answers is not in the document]
jeffkaufman: See https://github.com/naobservatory/p2ra/pull/7 for where I fixed this for `hsv_1.py`

pathogens/hva.py
jeffkaufman: Done in https://github.com/naobservatory/p2ra/pull/10

pathogens/hva.py
jeffkaufman: Done in https://github.com/naobservatory/p2ra/pull/10

pathogens/hva.py
jeffkaufman: Also should include confidence interval

simonleandergrimm: https://www.cdc.gov/hepatitis/statistics/2018surveillance/HepA.htm

pathogens/hva.py
simonleandergrimm: (Not actually outdated, I still need to change the source to the census.

pathogens/hva.py
```diff
@@ -0,0 +1,100 @@
+import sys
+
+sys.path.append("..")
+
+from p2ra_prevalences.pathogen_properties import *
+import math
+
+background = """Hepatitis A is a vaccine-preventable liver-infection caused by the hepatitis A virus. In the US,
+it's mostly spread by individual contact. There is little seasonal variance in Hepatitis A incidence (https://www.cdc.gov/vaccines/pubs/pinkbook/hepa.html). Viral shedding persists for 1 to 3 weeks."""
```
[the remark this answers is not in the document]
jeffkaufman: This is 112 columns though?

pathogens/hva.py
jeffkaufman: I don't see where you've adjusted for underreporting?

I expect when you do you'll want to use `Scalar` introduced in #12

pathogens/hva.py
jeffkaufman: I still don't see this?

jeffkaufman (changes requested):

pathogens/hva.py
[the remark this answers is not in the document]
simonleandergrimm: Fixed

pathogens/hva.py
simonleandergrimm: I added the variables king_county_absolute_2017 and king_county_absolute_2018.

pathogens/hva.py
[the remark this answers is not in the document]
jeffkaufman: Still not fixed

pathogens/hva.py
```diff
@@ -0,0 +1,101 @@
...
+)
+
+
+us_incidence_absolute_2018 = IncidenceAbsolute(
+    annual_infections=12474,
+    confidence_interval=(17500, 27400),
+    country="United States",
+    source="https://www.cdc.gov/hepatitis/statistics/2018surveillance/HepA.htm",
+    start_date="2018-01-01",
+    end_date="2018-12-31",
```
jeffkaufman, line 23: This (and all below) can now just be `date="YYYY"`.  So `date="2018"`.

jeffkaufman (approved):
Looks good!  Feel free to go ahead and merge after handling the two small comments below.

pathogens/hva.py
```diff
@@ -0,0 +1,98 @@
...
+    taxid=208726,
+)
+
+
+us_incidence_absolute_2018 = IncidenceAbsolute(
+    annual_infections=12474,
+    confidence_interval=(17500, 27400),
+    country="United States",
+    source="https://www.cdc.gov/hepatitis/statistics/2018surveillance/HepA.htm",
+    date="2018",
```
jeffkaufman, line 22: Just noticed that this won't work until #15 goes in, sorry!  But #15 should be going in as soon as you finish your review...

PR #11, opened 2023-04-20
Add requirements files

dp-rice:
I made one file for real requirements and "requirement-dev.txt" for mypy, black, etc. I also updated the presubmit hook to use the file. The second commit is a typo fix and should be squashed.


dp-rice:
Moved line length to a pyproject.toml config file so that it's the same when I run black locally to fix my file and when it runs to check.

.github/workflows/presubmit.yml
```diff
@@ -8,5 +8,5 @@ jobs:
         with:
             python-version: '3.11'
       - uses: actions/checkout@v3
-      - run: python3 -m pip install mypy black
+      - run: python3 -m pip install -r requirements-dev.txt
```
jeffkaufman, line 11: You'll need `requirements.txt` here too once you add any tests that depend on `pystan`, though I guess it's fine to wait until then to sort this out.

jeffkaufman (approved):

.github/workflows/presubmit.yml
dp-rice: I decided to add requirements.txt to requirements-dev.txt. It's unlikely we'll ever want the dev requirements alone.

dp-rice:
I don't mind rolling back `isort` if you don't care for it, but mypy, black, isort are my usual trio for python projects.

jeffkaufman (approved):

PR #14, opened 2023-04-21
format: add script that formats all python code in the repo

jeffkaufman:


jeffkaufman:
This isn't needed after #11 added `pyproject.toml`.  Now you can just run `black`

PR #22, opened 2023-04-24
Hsv 2

simonleandergrimm:


pathogens/hsv_2.py
```diff
@@ -0,0 +1,63 @@
...
+persist lifelong (CDC, https://www.hopkinsmedicine.org/health/
+conditions-and-diseases/herpes-hsv1-and-hsv2), manifesting as cold sores or
+fever blisters (https://www.hopkinsmedicine.org/health/conditions-and-diseases/
+herpes-hsv1-and-hsv2). After initial infection and potential symptoms, most
+HSV-1 infections persist lifelong."""
+
+
+pathogen_chars = PathogenChars(
+    na_type=NAType.DNA,
+    enveloped=Enveloped.NON_ENVELOPED,
```
jeffkaufman, line 14: I thought HSV was enveloped?

pathogens/hsv_2.py
```diff
@@ -0,0 +1,63 @@
...
+    taxid=10310,
+)
+
+
+cdc_2015_2016_nhanes_seroprevalence = Prevalence(
+    infections_per_100k=0.102 * 100_000,
```
jeffkaufman, line 20: Where does 0.102 come from?  I think this should be `infections_per_100k= 478/2813 * 100_000`?

pathogens/hsv_2.py
```diff
@@ -0,0 +1,63 @@
...
+    number_of_participants=3710,
```
jeffkaufman, line 21: What would you think of `2813` (positive + negative) since 3710 includes the indeterminate and missing results?

pathogens/hsv_2.py
```diff
@@ -0,0 +1,63 @@
...
+    country="United States",
+    start_date="2015",
+    end_date="2016",
+    source="https://wwwn.cdc.gov/Nchs/Nhanes/2015-2016/HSV_I.htm#:~:text=3710-,LBXHE2%20%2D%20Herpes%20Simplex%20Virus%20Type%202,-Variable%20Name%3A,",
+)
+
+
+cdc_2018_nhanes_estimate = PrevalenceAbsolute(
+    infections=18.6 * 1e6,
+    confidence_interval=(18.1 * 1e6, 19.0 * 1e6),
```
jeffkaufman, line 31: track that 18.1 and 19.0 represent the first and third quartiles ("We used Monte Carlo simulation to generate 10,000 input parameter sets for age and sex subpopulations and summarized our estimates by their median; uncertainty intervals for these estimates are characterized by their first (Q1) and third (Q3) quartiles. ...  Q1 = 18.1, Q3 = 19.0")

pathogens/hsv_2.py
```diff
@@ -0,0 +1,63 @@
...
+    date="2018",
```
jeffkaufman, line 32: The analysis is 2018, but the data is older: "Using NHANES data from the 2011–2014 and 2015–2018 NHANES cycles..."

pathogens/hsv_2.py
```diff
@@ -0,0 +1,63 @@
...
+    country="United States",
+    confidence_interval=(0.0966 * 100_000, 0.1495 * 100_000),
+    start_date="2015",
+    end_date="2016",
+    source="https://www.cdc.gov/nchs/data/databriefs/db304_table.pdf?#page=3",
+    methods="https://www.cdc.gov/nchs/products/databriefs/db304.htm#:~:text=Estimates%20were%20calculated,p%20%3C%200.05.",
+)
+
+
+us_population_2018 = Population(
```
jeffkaufman, line 50: This is the wrong population: "We estimated a total of 18.6 (Q1 = 18.1, Q3 = 19.0) million prevalent genital herpes infections in 2018 among 18- to 49-year-olds overall caused by HSV-2".  You need to divide 18.6M by the number of 18-49yo Americans.

jeffkaufman (changes requested):

pathogens/hsv_2.py
simonleandergrimm: In clinical trials you often use method where you use drop-outs "against" you: https://en.wikipedia.org/wiki/Intention-to-treat_analysis

I have no strong feelings about this, let me know what you prefer

pathogens/hsv_2.py
simonleandergrimm: will fix

pathogens/hsv_2.py
simonleandergrimm: In clinical trials you often use method where you use drop-outs "against" you: https://en.wikipedia.org/wiki/Intention-to-treat_analysis

I have no strong feelings about this, let me know what you prefer

jeffkaufman: I don't think that makes sense here:

* In a clinical trial drop-outs could be casually downstream from what you're trying to test
* In a clinical trial you're voluntarily taking on a higher burden of proof
* Erring on the low side when estimating `infections_per_100k` isn't "against us".  If anything it's "for us" because it gives a more optimistic estimate of the feasibility of wastewater-based detection

pathogens/hsv_2.py
```diff
@@ -0,0 +1,100 @@
...
+
+us_population_2018_18_to_49yo = Population(
+    people=stratified_us_pop["15-19"].people * 0.6
+    + stratified_us_pop["20-24"].people
+    + stratified_us_pop["25-34"].people
+    + stratified_us_pop["35-44"].people
+    + stratified_us_pop["45-54"].people * 0.5,
+    date="2018",
+    country="United States",
+    source="https://data.census.gov/table?q=2018+us+population&t=Civilian+Population",
```
jeffkaufman, line 91: Instead of a `source`, this should have `inputs=list(stratified_us_pop.values())`

pathogens/hva.py
```diff
@@ -17,6 +17,7 @@
 us_incidence_absolute_2018 = IncidenceAbsolute(
     annual_infections=12474,
     confidence_interval=(17500, 27400),
+    coverage_probability=0.95,
```
jeffkaufman, line 20: These `coverage_probability` additions, here and below, belong in a separate PR

jeffkaufman (changes requested):

pathogens/hva.py
jeffkaufman: I see, you did them in #23 -- the issue is that you should have either (a) waited to merge #23 until after #22 was in (at which point you merge it to `main`) or (b) set up #23 directly as a pull relative to `main` that only included adding `coverage_probability`.

Not worth fixing now, but something to keep in mind for next time you're chaining PRs!

PR #23, opened 2023-04-24
Added ability to give confidence interval coverage.

simonleandergrimm:
Added said coverage, where applicable, to hsv_2, hva, and sars_cov_2. Named variable based on this article: https://en.wikipedia.org/wiki/Coverage_probability https://en.wikipedia.org/wiki/Coverage_probabilityhttps://en.wikipedia.org/wiki/Coverage_probability


pathogen_properties.py
```diff
@@ -42,6 +42,7 @@ class Variable:
     county: Optional[str] = None
     number_of_participants: Optional[int] = None
     confidence_interval: Optional[tuple[float, float]] = None
+    coverage_probability_perc_point: Optional[float] = None
```
jeffkaufman, line 45: I think this is clearer as just a fraction: `confidence_interval_coverage=0.5`, `confidence_interval_coverage=0.95`; what do you think?

jeffkaufman (changes requested):

pathogen_properties.py
simonleandergrimm: I thought about this and have no strong opinion. Will change to fraction.

PR #32, opened 2023-04-25
Taxids

### LINKED ISSUE #28: Typesafe TaxIDs
Currently TaxIDs are represented as ints in the prevalence code and strings in the MGS data. This makes it easy to make mistakes, e.g. when searching a dict. Semantically TaxIDs are neither (it doesn't make sense to do arithmetic or string operations on them). I propose to make a new datatype `TaxID = NewType("TaxID", int)` so we can use the type system to check their usage.

dp-rice:
Resolves #28. There are two versions here in successive commits. The first commit does the simple thing of representing TaxID as a NewType of int. The second commit changes it to a datatype so that it can be initialized with (appropriate) strings as well. @jeffkaufman do you have a preference?

### DIFF
--- a/pathogen_properties.py
+++ b/pathogen_properties.py
@@ -1,7 +1,7 @@
 import os.path
 from dataclasses import dataclass
 from enum import Enum
-from typing import Optional
+from typing import NewType, Optional
 
 # Enums, short for enumerations, are a data type in Python used to represent a set of named values,
 # which are typically used to define a set of related constants with unique names.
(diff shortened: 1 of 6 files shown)


jeffkaufman:
[0c9b47e](https://github.com/naobservatory/p2ra/pull/32/commits/0c9b47ec06296c7591eb6fdfc37ee729beb9d85e) seems good.

With the fancy version, I'm not sure it's worth it relative to `TaxID(int(stringly_taxid_from_file))`? Unless `__post_init__` could let you do something like:

```
pathogen_chars = PathogenChars(
    na_type=NAType.RNA,
    enveloped=Enveloped.NON_ENVELOPED,
    taxid=208726,
)
```

Would that be possible?

dp-rice:
I don't think so. That would either require runtime type inference which Python doesn't do, or a post-init in PathogenChars, which I don't think is the right place for the type casting to happen.
To me the main benefit of the complicated version is that it allows a custom error message when you pass invalid data. (Also just realized that my error message doesn't report the invalid value so I'll fix that if we decide to keep it).
If you don't think that's worth the extra lines of code, I'll revert to the simple version.

jeffkaufman (approved):

PR #38, opened 2023-04-25
Pull in read count data from the MGS dashboard repo

dp-rice:
Adds the ability to pull BioProject and Sample metadata; taxonomic trees; and human viral read counts from the [MGS pipeline repo](https://github.com/naobservatory/mgs-pipeline/tree/main/dashboard).
Defines two modules and one script:

- `tree.py` defines a generic Tree class to represent data distributed on taxonomic trees.
- `mgs.py` defines utilities for getting and manipulating data from the pipeline repo
- `get_rothman_virus_counts.py` demonstrates getting the data and prints the read counts for each pathogen in each sampling location the Rothman dataset. Its output looks like:

```bash
> ./get_rothman_virus_counts.py
sars_cov_2
 All	408
 SB	0
 OC	32
 SJ	29
 HTP	329
 PL	5
 ESC	12
 JW	0
 NC	1
 JWPCP	0
hiv
 All	1
 SB	0
 OC	1
 SJ	0
 HTP	0
 PL	0
 ESC	0
 JW	0
 NC	0
 JWPCP	0
hsv_1
 All	41
 SB	2
 OC	6
 SJ	2
 HTP	12
 PL	9
 ESC	1
 JW	0
 NC	6
 JWPCP	3
norovirus
 All	1636
 SB	385
 OC	169
 SJ	171
 HTP	735
 PL	139
 ESC	11
 JW	1
 NC	22
 JWPCP	3
hav
```


dp-rice:
@jeffkaufman I'll probably reorganize this code and I still need to write tests, but it would be helpful if you could take a look and see whether it looks like what I'm doing makes sense. One difference from the dashboard is that I am counting reads attributed to non-leaf nodes of the taxonomy. Do you think that's the right thing to do, or should I try to match the dashboard exactly?

get_mgs_data.py
```diff
@@ -0,0 +1,153 @@
...
+
+from pathogen_properties import TaxID
+from pathogens import pathogens
+
+Sample = NewType("Sample", str)
+
+
+def load_samples(mgs_dir: Path, bioproject: str) -> list[Sample]:
+    with open(
+        mgs_dir / "dashboard/metadata_bioprojects.json"
```
jeffkaufman, line 20: I haven't used pathlib before, but this is neat!  I should probably switch to it.

get_mgs_data.py
```diff
@@ -0,0 +1,153 @@
...
+    ) as projects_file:
+        data = json.load(projects_file)
+    return [Sample(s) for s in data[bioproject]]
+
+
+class SampleAttributes(BaseModel):
+    country: str
+    location: str
+    fine_location: str | None = None
```
jeffkaufman, line 29: I thought earlier you said you preferred `Optional[...]` to `... | None`?

dp-rice: You're right! Don't know why I did that.

get_mgs_data.py
```diff
@@ -0,0 +1,153 @@
...
+    def __iter__(self):
+        yield self
+        for child in self.children:
+            yield from child
+
+    def __getitem__(self, val: T) -> Tree[T] | None:
+        return get_subtree(self, val)
+
+
+def get_subtree(tree: Tree[T], val: T) -> Tree[T] | None:
```
jeffkaufman, line 82: I'd put this and the other tree operations inside Tree

jeffkaufman:
> One difference from the dashboard is that I am counting reads attributed to non-leaf nodes of the taxonomy. Do you think that's the right thing to do, or should I try to match the dashboard exactly?

That was a dashboard bug; thanks for catching it!  I've fixed this in 022f461.

I think your code and the dashboard should give the same numbers, but if the dashboard is wrong that's not necessarily a problem with your code ;)

dp-rice:
I'm working on a version that grabs the json files from github rather than reading them from disk. I'm thinking that this will facilitate testing and also make it simpler to use. Does that seem like a good idea?

jeffkaufman:
> I'm working on a version that grabs the json files from github rather than reading them from disk. I'm thinking that this will facilitate testing and also make it simpler to use. Does that seem like a good idea?

Whichever feels more comfortable to you!

dp-rice:
I think this is ready to merge. The next step will be to combine this data with the pathogen data and run the statistics on it.
I'm not super happy with the tests I wrote. They're a combination of fuzzing and a few arbitrary simple examples. I'd appreciate any advice on writing better tests. Also, what do you think of property-based testing (e.g. [hypothesis](https://hypothesis.readthedocs.io/en/latest/)). I like it in principle but the tests take a long time to run.

get_mgs_data.py
```diff
@@ -0,0 +1,31 @@
+import mgs
```
jeffkaufman, line 1: `#!/usr/bin/env python3`

get_mgs_data.py
```diff
@@ -0,0 +1,31 @@
...
+from pathogens import pathogens
+
+repo = mgs.GitHubRepo(user="naobservatory", repo="mgs-pipeline", branch="main")
```
jeffkaufman, line 4: Code that actually does something should be downstream from an `if __name__ == "__main__"`

get_mgs_data.py
```diff
@@ -0,0 +1,31 @@
...
+bioproject = mgs.BioProject("PRJNA729801")  # Rothman
+samples = bp_data[bioproject]
+sample_attribs = {s: sample_data[s] for s in samples}
+fine_locs = set(sample_attribs[s].fine_location for s in samples)
+
+for pathogen_name, pathogen in pathogens.items():
+    print(pathogen_name)
+    taxid = pathogen.pathogen_chars.taxid
+    subtree = taxtree[taxid]
+    if subtree:
```
jeffkaufman, line 19: nit: I generally prefer `if not subtree: continue` to avoid deep indentation and make it clear there's nothing hiding later, but up to you

get_mgs_data.py
```diff
@@ -0,0 +1,31 @@
...
+counts = mgs.load_sample_counts(repo)
+taxtree = mgs.load_tax_tree(repo)
+
+bioproject = mgs.BioProject("PRJNA729801")  # Rothman
+samples = bp_data[bioproject]
+sample_attribs = {s: sample_data[s] for s in samples}
+fine_locs = set(sample_attribs[s].fine_location for s in samples)
+
+for pathogen_name, pathogen in pathogens.items():
+    print(pathogen_name)
```
jeffkaufman, line 16: Can you include the output of this script in the PR description?  Generally good to do this anytime you're adding something that prints things, as a quick substitute for writing tests.

mgs.py
```diff
@@ -0,0 +1,104 @@
...
+from dataclasses import dataclass, field
+from datetime import date
+from typing import NewType, Optional
+
+from pydantic import BaseModel
+
+from pathogen_properties import TaxID
+from tree import Tree, tree_from_list
+
+repo_url = "https://raw.githubusercontent.com/naobservatory/mgs-pipeline/main/"
```
jeffkaufman, line 13: unused

mgs.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+@dataclass
+class GitHubRepo:
+    user: str
+    repo: str
+    branch: str
+    url: str = field(init=False)
```
jeffkaufman, line 21: Since this is only used in one place (the beginning of `get_file`) I think it's not worth the `__post_init__` complexity.  You can just start `get_file` with `file_url = f"https://raw.githubusercontent.com/{self.user}/{self.repo}/{self.branch}/{path}`

tree.py
```diff
@@ -0,0 +1,51 @@
...
+            return None
+
+    def to_list(self) -> list:
+        return [self.data] + [c.to_list() for c in self.children]
+
+    def map(self, f: Callable[[T], S]) -> Tree[S]:
+        return Tree(f(self.data), [c.map(f) for c in self.children])
+
+
+def tree_from_list(input: list) -> Tree:
```
jeffkaufman, line 50: I think this would be clearer as a `@staticmethod` on Tree?  So `Tree.from_list(l)` and not `tree_from_list(l)`?

jeffkaufman (changes requested):
Looks good!  Only minor issues

jeffkaufman:
Also, PR description should say what this PR actually does.

tree.py
dp-rice: Good catch

mgs.py
dp-rice: Good point. I had originally thought we might want do do other stuff with this, but that seems unlikely.

get_rothman_virus_counts.py
```diff
@@ -0,0 +1,37 @@
...
+    fine_locs = set(sample_attribs[s].fine_location for s in samples)
+
+    for pathogen_name, pathogen in pathogens.items():
+        print(pathogen_name)
+        taxid = pathogen.pathogen_chars.taxid
+        subtree = taxtree[taxid]
+        if not subtree:
+            continue
+        virus_counts = mgs.count_reads(subtree, counts)
+        print("All", sum(virus_counts[s] for s in samples), sep="\t")
```
jeffkaufman, line 27: nit: I'd indent this (`" All"`)

get_rothman_virus_counts.py
```diff
@@ -0,0 +1,37 @@
...
+        for fine_loc in fine_locs:
+            print(
```
jeffkaufman, line 29: This too

test.py
```diff
@@ -32,5 +35,89 @@ def test_properties_exist(self):
...
+    def test_map_composition(self):
+        f = lambda x: x + 1
+        g = lambda x: 2 * x
+        self.assertEqual(
+            self.node.map(f).map(g), self.node.map(lambda x: g(f(x)))
+        )
+        self.assertEqual(
+            self.leaf.map(f).map(g), self.leaf.map(lambda x: g(f(x)))
+        )
+
```
jeffkaufman, line 120: It would be nice to have a test that verifies this matches [0, [1], [2], [3]]

tree.py
```diff
@@ -0,0 +1,53 @@
...
+T = TypeVar("T")
+S = TypeVar("S")
+
+
+@dataclass
+class Tree(Generic[T]):
+    data: T
+    children: list[Tree[T]] = field(default_factory=list)
+
+    def _helper(self, depth: int) -> str:
```
jeffkaufman, line 15: `_str_helper`

jeffkaufman (approved):
looks good, only minor comments

(approving now to indicate I don't need to re-review after you fix these unless something surprising comes up)

PR #44, opened 2023-04-27
Use MGS data to fit model

### LINKED ISSUE #41: Complete inference example for one pathogen
We should pick one pathogen and one MGS study and run the analysis start-to-finish (pulling prevalence and MGS data and doing the stats) as a prototype. The best candidate is SARS-CoV2 in the Rothman data: there are plenty of counts, detailed prevalence estimates, and we can compare with Jeff's [earlier analysis](https://www.jefftk.com/p/case-rates-to-sequencing-reads).

dp-rice:
Resolves #41 (but using Norovirus instead of SARS-CoV2 for simplicity). Creates an `MGSData` object that provides a standard interface to the MGS data for the statistical fitting scripts to call.

```
> ./fit_norovirus.py 2> /dev/null  # discard Stan status output
------------------------------------------------
Relative abundance of norovirus at 1% prevalence
------------------------------------------------
Naive estimate:
0.00301
Posterior arithmetic mean:
0.00420
Posterior geometric mean:
0.00257
Posterior quantiles:
5%		25%		50%		75%		95%
0.00055		0.00129		0.00256		0.00490		0.01314

### CODE REVIEW

File: test.py [hunk 1]
```diff
@@ -79,6 +76,32 @@ def test_count_reads(self):
         self.assertEqual(mgs.count_reads(taxtree, sample_counts), expected)
 
 
+class TestMGSData(unittest.TestCase):
+    # TODO: Fake this?
```
jeffkaufman, line 80: I don't see a reason to fake this, fwiw

jeffkaufman (approved):
Looks reasonable!

I haven't reviewed this super closely; let me know if you want more than a read through, but I think that's not necessary at this stage?

PR #47, opened 2023-04-27
Added an Enum to keep track of whether an estimate is for active shedding or latent prevalence.

asherps:
Added an Enum to keep track of whether an estimate is for active shedding or latent prevalence. It seems important to clarify whether a given estimate is for what percentage of the population is shedding at a given time or for what percent has the latent disease at a given time.  It's also useful to be able to create both an active and a latent estimate for one virus and be able to distinguish between them.


pathogen_properties.py
```diff
@@ -218,6 +229,7 @@ def to_prevalence(self, shedding_duration: SheddingDuration) -> Prevalence:
             * shedding_duration.days
             / 365,
             inputs=[self, shedding_duration],
+            active=Active.ACTIVE,
```
jeffkaufman, line 232: Can you add a comment explaining why any prevalence derived from an incidence and a shedding duration should be seen as an estimate of active infections?

jeffkaufman (approved):
Looks good!  Just needs the one comment in `to_prevalence`

pathogen_properties.py
simonleandergrimm: @jeffkaufman Wouldn't the right state of this review be in "Request changes," not approval?

jeffkaufman: approval with comments means "these comments are minor enough that if you make the requested changes and don't run into any issues while doing so it's fine to merge without getting another review from me"

PR #50, opened 2023-04-28
check: add --fix option to automatically reformat code and sort imports

jeffkaufman:
This has some manual argument parsing.  I think this is fine since we're only doing one optional argument, though if we later add more we should switch to `getops` or something.

```
$ ./check.sh --fix
reformatted /Users/jeffkaufman/code/p2ra/pathogen_properties.py

All done! ✨ 🍰 ✨
1 file reformatted, 13 files left unchanged.
Fixing /Users/jeffkaufman/code/p2ra/pathogen_properties.py
Skipped 2 files
Testing...
................
----------------------------------------------------------------------
Ran 16 tests in 0.817s

OK
Running mypy to check types...
Success: no issues found in 14 source files
Running black to check formatting...
Running isort to check import sorting...
OK after fixes
```

And then if I run it again:

```
$ ./check.sh --fix
Testing...
................
----------------------------------------------------------------------
Ran 16 tests in 2.375s

OK
Running mypy to check types...
Success: no issues found in 14 source files
Running black to check formatting...
Running isort to check import sorting...
OK
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,9 @@ Before creating a PR or submitting code, run `./check.sh`.  It will run tests
 than check your types and formatting.  This also runs automatically on GitHub
 when you make PRs, but it's much faster to catch problems locally first.
 
+If `./check.sh` complains about formatting or import sorting, you can fix this
+automatically with `./check.sh --fix`.
+
 #### Installing pystan
 
 Pystan should be installed along with the other requirements when you run:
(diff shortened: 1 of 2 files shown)


dp-rice (approved):
Looks good! My only suggestion would be to make the output a little louder if the fixing scripts make changes, so that people don't miss it and forget to commit the changes. When I ran it with incorrect files I was confused for a moment because the checks at the end say "OK". Maybe you could capture their output and emit it at end of the script.

dp-rice:
Looks good. Thanks!

dp-rice (approved):

PR #53, opened 2023-04-28
norovirus: more detailed prevalence estimate

jeffkaufman:
Previously I did some manual calculations to see how the Rothman period compared to normal, but this throws away a lot of data that could be useful in modeling.  Now I generate two estimates for every month, one using national outbreaks and one using CA-only outbreaks.

I'm not sure we actually want the CA-only ones, though?  They're kind of noisy, and do we expect LA county to follow CA-overall much more closely than national overall?

Before we were estimating 0.82 per 100k for the Rothman data.  With this PR we estimate:

```
National:
  2020-08 | 0.33
  2020-09 | 0.51
  2020-10 | 0.66
  2020-11 | 0.85
  2020-12 | 0.98
  2021-01 | 4.92

CA:
  2020-08 | 0
  2020-09 | 0
  2020-10 | 9.45
  2020-11 | 0
  2020-12 | 0
  2021-01 | 0
```

The CA noise is that we have a single outbreak in October 2020, and that's enough to move our estimate a lot.

### DIFF
--- a/pathogens/norovirus.py
+++ b/pathogens/norovirus.py
@@ -1,3 +1,7 @@
+from collections import Counter, defaultdict
+
+import numpy as np
+
 from pathogen_properties import *
 
 background = """Norovirus in a GI infection, mostly spread through personal
@@ -11,7 +15,7 @@ pathogen_chars = PathogenChars(
 )
 
 
-cases = IncidenceAbsolute(
+normal_year_national_cases = IncidenceAbsolute(
     annual_infections=20e6,
     confidence_interval=(19e6, 21e6),
     country="United States",
@@ -33,25 +37,94 @@ shedding_duration = SheddingDuration(
     source="https://www.mayoclinic.org/diseases-conditions/norovirus/symptoms-causes/syc-20355296",
 )
 
-rothman_period_outbreaks = Number(
-    number=13 / 5,  # 13 outbreaks over 5 months
-    country="United States",
-    source="https://www.cdc.gov/norovirus/reporting/norostat/data-table.html",
-    start_date="2020-08-01",
-    end_date="2020-12-31",
-)
-normal_year_outbreaks = Number(
-    number=1246 / 12,  # 1246 outbreaks over one year
-    country="United States",
-    source="https://www.cdc.gov/norovirus/reporting/norostat/data-table.html",
-    start_date="2012",
-    end_date="2022",
-)
+
+def days_in_month(date):
+    _, last_day = calendar.monthrange(date.year, date.month)
+    return last_day
 
 
 def estimate_prevalences():
-    return [
-        cases.to_rate(us_population)
-        .to_prevalence(shedding_duration)
-        .scale(rothman_period_outbreaks.per(normal_year_outbreaks))
-    ]
+    us_outbreaks = Counter()  # date -> count
+    ca_outbreaks = Counter()  # date -> count
+
+    prevalences = []
+
+    # Downloaded from https://wwwn.cdc.gov/norsdashboard/ 2023-04-28.
+    with open(prevalence_data_filename("cdc-nors-outbreak-data.tsv")) as inf:
+        cols = None
+        for line in inf:
+            row = line.strip().split("\t")
+            if not cols:
+                cols = row
+                continue
+
+            year = int(row[cols.index("Year")])
+            month = int(row[cols.index("Month")])
+            state = row[cols.index("State")]
+            etiology = row[cols.index("Etiology")]
+            genotype = row[cols.index("Serotype or Genotype")]
+            if "Norovirus" not in etiology:
+                continue
+
+            date = year, month
+
+            us_outbreaks[date] += 1
+            if state == "California":
+                ca_outbreaks[date] += 1
+
+    normal_year_national_prevalence = normal_year_national_cases.to_rate(
+        us_population
+    ).to_prevalence(shedding_duration)
+
+    total_us_outbreaks = 0
+    total_ca_outbreaks = 0
+    days_considered = 0
+    for year in range(2012, 2020):
+        for month in range(1, 13):
+            total_us_outbreaks += us_outbreaks[year, month]
+            total_ca_outbreaks += ca_outbreaks[year, month]
+            days_considered += days_in_month(datetime.date(year, month, 1))
+    normal_us_average_daily_outbreaks = total_us_outbreaks / days_considered
+    normal_ca_average_daily_outbreaks = total_ca_outbreaks / days_considered
+
+    for target_year in range(2012, 2022):
+        for target_month in range(1, 13):
+            target_date = "%s-%s" % (target_year, str(target_month).zfill(2))
+
+            target_us_daily_outbreaks = us_outbreaks[
+                target_year, target_month
+            ] / days_in_month(datetime.date(target_year, target_month, 1))
+            target_ca_daily_outbreaks = ca_outbreaks[
+                target_year, target_month
+            ] / days_in_month(datetime.date(target_year, target_month, 1))
+
+            prevalences.append(
+                normal_year_national_prevalence.scale(
+                    Scalar(
+                        scalar=target_us_daily_outbreaks
+                        / normal_us_average_daily_outbreaks,
+                        country="United States",
+                        state="California",
+                        date=target_date,
+                        source="https://wwwn.cdc.gov/norsdashboard/",
+                    )
+                ).target(country="United States", date=target_date)
+            )
+
+            prevalences.append(
+                normal_year_national_prevalence.scale(
+                    Scalar(
+                        scalar=target_ca_daily_outbreaks
+                        / normal_ca_average_daily_outbreaks,
+                        country="United States",
+                        state="California",
+                        date=target_date,
+                        source="https://wwwn.cdc.gov/norsdashboard/",
+                    )
+                ).target(
+                    country="United States",
+                    state="California",
+                    date=target_date,
+                )
+            )
+    return prevalences


jeffkaufman:
(I'm going to ask Dan for review after Simon is done)

pathogens/norovirus.py
```diff
@@ -33,25 +37,102 @@
...
+            if not cols:
+                cols = row
+                continue
+
+            year = int(row[cols.index("Year")])
+            month = int(row[cols.index("Month")])
+            state = row[cols.index("State")]
+            etiology = row[cols.index("Etiology")]
+            # It distinguishes between GI and GII Norovirus.  I'm currently
+            # discarding this, but it could potentially be useful?
```
simonleandergrimm, line 66: I think it could be useful as we could check how variable metagenomic wastewater measurements are between G1 and G2 Norovirus. I face the same question with HPV subtypes, we have NHANES cohort PCR positivity rates for a lot of different HPV types. 

We likely have ways to easily handle more fine-grained data, which makes me lean to include data on GI and GII here.

pathogens/norovirus.py
```diff
@@ -11,7 +15,7 @@
 )
 
 
-cases = IncidenceAbsolute(
+normal_year_national_cases = IncidenceAbsolute(
     annual_infections=20e6,
     confidence_interval=(19e6, 21e6),
```
simonleandergrimm, line 20: Do we know from which year this estimate is?

pathogens/norovirus.py
```diff
@@ -33,25 +37,102 @@
...
+                target_year, target_month
+            ] / days_in_month(datetime.date(target_year, target_month, 1))
+
+            prevalences.append(
+                normal_year_national_prevalence.scale(
+                    Scalar(
+                        scalar=target_us_daily_outbreaks
+                        / normal_us_average_daily_outbreaks,
+                        country="United States",
+                        state="California",
```
simonleandergrimm, line 114: This estimate is national, not state-specific, so doesn't require state-specific location details

pathogens/norovirus.py
```diff
@@ -33,25 +37,102 @@
...
+                        date=target_date,
+                        source="https://wwwn.cdc.gov/norsdashboard/",
+                    )
+                ).target(country="United States", date=target_date)
+            )
+
+            # The CA-specific ones are very noisy; consider dropping them?
+            prevalences.append(
```
simonleandergrimm, line 122: Can you provide information on how they are noisy? Or is there an easy way for me to assess their noisiness myself?

simonleandergrimm (changes requested):

pathogens/norovirus.py
jeffkaufman: Have a look at the examples in the PR description?  A single CA outbreak is enough to massively shift our predictions.

pathogens/norovirus.py
jeffkaufman: Turns out this comes from https://wwwnc.cdc.gov/eid/article/19/8/pdfs/13-0465.pdf which is a much better source and gets into the details of how they got the estimate.  I haven't dug into the details, but I've switched the source link.

pathogens/norovirus.py
```diff
@@ -33,25 +37,102 @@
     source="https://www.mayoclinic.org/diseases-conditions/norovirus/symptoms-causes/syc-20355296",
```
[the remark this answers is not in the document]
jeffkaufman: Added!

pathogens/norovirus.py
```diff
@@ -33,25 +37,102 @@
...
-        cases.to_rate(us_population)
-        .to_prevalence(shedding_duration)
-        .scale(rothman_period_outbreaks.per(normal_year_outbreaks))
-    ]
+    us_outbreaks = Counter()  # date -> count
+    ca_outbreaks = Counter()  # date -> count
+
+    prevalences = []
+
+    # Downloaded from https://wwwn.cdc.gov/norsdashboard/ 2023-04-28.
```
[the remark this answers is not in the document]
jeffkaufman: Fixed!

pathogens/norovirus.py
jeffkaufman: This is a bit tricky; raised https://twist.com/a/197793/ch/661244/t/4356084/ for more discussion

pathogens/norovirus.py
jeffkaufman: good catch; thanks!

pathogens/norovirus.py
simonleandergrimm: Right, I missed that! Dan probably has more useful things to say, based on our conversation in the check-in I would think that we can adopt the US estimates for California, shift estimates down for the months that aren't 10-2021, and shift the estimate up for 10-2021. @dp-rice happy to hear what you think.

pathogens/norovirus.py
jeffkaufman: Doing this in https://github.com/naobservatory/p2ra/pull/60

pathogens/norovirus.py
simonleandergrimm: @jeffkaufman If you agree with my reasoning above, we could replace the estimate what I laid out above, adding https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3375761/ as a source. You could also instantiate the calculation `100 / 26 * 5.4*e6` by adding 100/26 as a scalar and 5,461,731 (Credible Interval 90%: 3,227,078–8,309,480) as a prevalence estimate.

simonleandergrimm: Aside from that change, the doc looks good to!

simonleandergrimm (changes requested):

pathogens/norovirus.py
jeffkaufman: Thanks for digging into this!  That makes sense.  Changed.

The paper also lists a few other viruses, which might be useful to you?

pathogens/norovirus.py
jeffkaufman: Planning to leave this as is for now; we can always delete or ignore the CA estimate.

simonleandergrimm (approved):

simonleandergrimm:
Approved the new version and added Dan for the next review.

pathogens/norovirus.py
dp-rice: Agreed. It's probably not worth using, but I don't see any harm in keeping it.

pathogens/norovirus.py
```diff
@@ -11,47 +15,141 @@
...
+
+    total_us_outbreaks = 0
+    total_ca_outbreaks = 0
+    days_considered = 0
+    for year in range(2012, 2020):
+        for month in range(1, 13):
+            total_us_outbreaks += us_outbreaks[year, month]
+            total_ca_outbreaks += ca_outbreaks[year, month]
+            days_considered += days_in_month(datetime.date(year, month, 1))
+    normal_us_average_daily_outbreaks = total_us_outbreaks / days_considered
```
dp-rice, line 112: I think better to call this just `us_average_daily_outbreaks`. "normal" to me implies typicality, more like "modal".

pathogens/norovirus.py
```diff
@@ -11,47 +15,141 @@
...
-    start_date="2012",
-    end_date="2022",
-)
+
+def days_in_month(date):
+    _, last_day = calendar.monthrange(date.year, date.month)
+    return last_day
 
 
 def estimate_prevalences():
```
dp-rice, line 61: The logic of the estimates is being obscured by the number of different things going on in this function. I think it would be better to separate:

1. Reading data from files
2. Counting outbreaks
3. Estimating the average cases per outbreak
4. Estimating the cases per month from outbreaks and average cases per outbreak

(Not wedded to any particular breakdown into functions).

dp-rice (changes requested):
Overall, looks good. The only comment that I would like to (quickly) review again is the one about separating out the different steps of `estimate_prevalences` to make the logic clearer.

Also, while I think this is a reasonably way to estimate expected prevalences from outbreak data, I'd like to flag that there's a lot of uncertainty here about things like: What constitutes an outbreak? What fraction of cases are attributable to outbreaks? What is the size distribution of outbreaks? If we decide to spend time improving this estimate, we'll want to try to get some information about those questions.

pathogens/norovirus.py
jeffkaufman: broke this up a bunch

pathogens/norovirus.py
```diff
@@ -11,47 +15,141 @@
...
-    return [
-        cases.to_rate(us_population)
+    us_outbreaks = Counter()  # date -> count
+    ca_outbreaks = Counter()  # date -> count
+
+    prevalences = []
+
+    # Downloaded on 2023-04-28 from https://wwwn.cdc.gov/norsdashboard/
+    # Data runs through the end of 2021.
```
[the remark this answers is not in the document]
jeffkaufman: added comment

pathogens/norovirus.py
```diff
@@ -11,47 +15,141 @@
...
+                # though, not much useful to us.
+                continue
+
+            date = year, month
+
+            us_outbreaks[date] += 1
+            if state == "California":
+                ca_outbreaks[date] += 1
+
+    normal_year_national_prevalence = (
```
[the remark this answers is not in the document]
jeffkaufman: It's not quite that: it's relative to the 2006 population, but it's not an estimate for 2006 in particular.

Changed "normal" to "pre-covid"

pathogens/norovirus.py
```diff
@@ -11,47 +15,141 @@
...
         .to_prevalence(shedding_duration)
-        .scale(rothman_period_outbreaks.per(normal_year_outbreaks))
-    ]
+        .scale(us_total_relative_to_foodborne_2006)
+    )
+
+    total_us_outbreaks = 0
+    total_ca_outbreaks = 0
+    days_considered = 0
+    for year in range(2012, 2020):
```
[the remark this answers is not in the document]
jeffkaufman: added comment

pathogens/norovirus.py
jeffkaufman: changed to pre-covid

pathogens/norovirus.py
```diff
@@ -11,47 +15,141 @@
...
+        for month in range(1, 13):
+            total_us_outbreaks += us_outbreaks[year, month]
+            total_ca_outbreaks += ca_outbreaks[year, month]
+            days_considered += days_in_month(datetime.date(year, month, 1))
+    normal_us_average_daily_outbreaks = total_us_outbreaks / days_considered
+    normal_ca_average_daily_outbreaks = total_ca_outbreaks / days_considered
+
+    for target_year in range(2012, 2022):
+        for target_month in range(1, 13):
+            target_date = "%s-%s" % (target_year, str(target_month).zfill(2))
```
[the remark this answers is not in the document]
jeffkaufman: that is clearer, thanks!

jeffkaufman:
@dp-rice @simonleandergrimm I removed the CA-specific estimates: they added a lot of complexity for something that wasn't very illuminating

dp-rice (approved):
Looks good, thanks!

PR #60, opened 2023-05-01
norovirus: make estimates specifically for GI and GII

jeffkaufman:
We want to be able to track the Norovirus subtypes more accurately.

Adds the idea that a prevalence can be for a taxid more specific than the overall one for a file.

Follow-up to https://github.com/naobservatory/p2ra/pull/53#discussion_r1181559803 and https://twist.com/a/197793/ch/661244/t/4356084/


pathogens/norovirus.py
```diff
@@ -91,6 +94,18 @@ def estimate_prevalences():
...
+            seen_II = False
+            for etiology in etiologies.split("; "):
+                if etiology.endswith("Norovirus Genogroup I"):
+                    seen_I = True
+                elif etiology.endswith("Norovirus Genogroup II"):
+                    seen_II = True
+            if seen_I and not seen_II:
+                us_outbreaks_I[date] += 1
+            elif seen_II and not seen_I:
+                us_outbreaks_II[date] += 1
```
dp-rice, line 107: This block raises some questions about the data:

- What does it mean if both I and II are seen in an outbreak? I can see not wanting to double count, but on the other hand it seems wrong to undercount as well.
- To what extent does each outbreak come with one and only one genogroup label?
- If some outbreaks have neither label, is it because it was attributed to some other subtype, or is it missing data?

I think that counting the subtypes is only really valuable if they're being labeled consistently in the dataset in a way that's transparent to us.

summarize.py
```diff
@@ -51,7 +51,15 @@ def start(pathogen_names):
             location = estimate.summarize_location()
 
             prevalence = "%.2f per 100k" % estimate.infections_per_100k
-            line = "%s (%s; %s)" % (prevalence.rjust(18), location, date)
+            taxid = ""
```
dp-rice, line 54: Should `taxid` default to the higher-level taxon?

pathogens/norovirus.py
```diff
@@ -7,11 +7,14 @@
 background = """Norovirus in a GI infection, mostly spread through personal
 contact."""
 
+NOROVIRUS = TaxID(142786)
+NOROVIRUS_GROUP_I = TaxID(122928)
```
dp-rice, line 11: It might be good to have an interface in the module to get all the TaxIDs present. Maybe an optional field in PathogenChars for `sub_taxids`.

dp-rice (approved):
I think my question about what fraction of outbreaks were labeled with a group I vs II was mostly answered by a comment, but it could be good to have a bit more info about how any outbreaks are neither/both.

pathogens/norovirus.py
jeffkaufman: > What does it mean if both I and II are seen in an outbreak?

It's possible, but rare: ~1%.  Added something to the code which calculates this and asserts that it's low enough to ignore.

> To what extent does each outbreak come with one and only one genogroup label?

40% of outbreaks are labeled, though it used to be ~100%: https://twist.com/a/197793/ch/661244/t/4356084/

> If some outbreaks have neither label, is it because it was attributed to some other subtype, or is it missing data?

The latter.  Added an assertion and a comment explaining.

summarize.py
jeffkaufman: It makes the output pretty noisy and is redundant with the headings, so I'd rather not.

pathogen_properties.py
```diff
@@ -152,13 +156,19 @@ def _location(self):
...
+
         return "; ".join(
             sorted(
                 set(i._location() for i in self.all_inputs if i._location())
             )
         )
 
     def summarize_date(self) -> Optional[tuple[datetime.date, datetime.date]]:
+        if self.is_target and self.parsed_start and self.parsed_end:
+            return self.parsed_start, self.parsed_end
```
jeffkaufman, line 170: Fixes a minor issue accidentally introduced in https://github.com/naobservatory/p2ra/pull/73

pathogens/norovirus.py
```diff
@@ -122,22 +127,46 @@ def load_nors_outbreaks() -> (
...
             if seen_I and not seen_II:
                 us_outbreaks_I[date] += 1
             elif seen_II and not seen_I:
                 us_outbreaks_II[date] += 1
 
+            if seen_other:
+                total_seen_other += 1
+
+            # We don't care about the I-vs-II labeling in old data.
+            if year >= HISTORY_START:
```
dp-rice, line 149: Why not also bounded by `HISTORY_END`?

dp-rice (approved):
👍

pathogens/norovirus.py
jeffkaufman: Hmm, this raises a good point about HISTORY_START and HISTORY_END not being parallel.  Renamed HISTORY_END to COVID_START.

We ignore after the start of covid in estimating norovirus patterns in normal times, because we want to compare with the overall pre-covid estimate.  But we so still pay attention to I-vs-II during covid because we want to check that our classifications are still good then.

PR #65, opened 2023-05-02
Asher vzv

asherps:
Completed estimates for latent VZV prevalence, active chickenpox prevalence, and active shingles prevalence. For the latent VZV estimate, I had to be loose with the formatting to make the estimate work, using Scalars a lot instead of more specific values.


pathogens/vzv.py
```diff
@@ -0,0 +1,119 @@
...
+population_above_34 = Prevalence(
+    infections_per_100k=0.55 * 100000,
+    date="2021",
+    active=Active.LATENT,
+    source="https://ourworldindata.org/grapher/median-age",
+)
+
+
+annual_new_VZV_infections = Scalar(
+    scalar=4000000,
```
simonleandergrimm, line 49: Structure long numbers with underscores: 4_000_000

.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+    "[python]": {
+        "editor.defaultFormatter": "ms-python.black-formatter"
+    },
+    "python.formatting.provider": "none"
+}
```
jeffkaufman, line 6: This should not be in this PR

pathogens/vzv.py
```diff
@@ -0,0 +1,119 @@
...
+    source="https://www.childrenshospital.org/conditions/chickenpox#:~:text=More%20than%2095%20percent%20of%20American%20adults%20have%20had%20chickenpox%20and%20about%204%2C000%2C000%20people%20get%20chickenpox%20every%20year",
+)
+
+US_Population = Population(
+    people=335000000,
```
simonleandergrimm, line 54: Would add a more formal source here, e.g. population projection for 2023 (appreciate that no Census data for 2023 seems to exist on data.census.gov)

simonleandergrimm: Also include seperators _

pathogens/vzv.py
```diff
@@ -0,0 +1,119 @@
...
+    tag="US-2023",
+    source="https://www.census.gov/popclock/",
+)
+
+prevalence_from_postvaccine_cases = Prevalence(
+    infections_per_100k=annual_new_VZV_infections.scalar
```
simonleandergrimm, line 60: I think this should be doable with the given toolkit from Jeff. Potentially it'd be good to have pathogen_properties.py open next to the pathogen doc, that is what I do to spot the appropriate functions.

pathogens/vzv.py
jeffkaufman, line 49: 4_000_000

pathogens/vzv.py
```diff
@@ -0,0 +1,119 @@
...
+    + prevalence_from_postvaccine_cases.infections_per_100k,
+    active=Active.LATENT,
+)
+
+
+# Now, calculations for active chickenpox and shingles estimates
+
+shingles_annual_infections = IncidenceAbsolute(
+    annual_infections=1000000,
+    tag="US-2023",
```
simonleandergrimm, line 79: This is from 2020. I would look into their references to get a better handle on the exact date of the source

pathogens/vzv.py
```diff
@@ -0,0 +1,119 @@
...
+    infections_per_100k=0.55 * 100000,
+    date="2021",
+    active=Active.LATENT,
+    source="https://ourworldindata.org/grapher/median-age",
+)
+
+
+annual_new_VZV_infections = Scalar(
+    scalar=4000000,
+    source="https://www.childrenshospital.org/conditions/chickenpox#:~:text=More%20than%2095%20percent%20of%20American%20adults%20have%20had%20chickenpox%20and%20about%204%2C000%2C000%20people%20get%20chickenpox%20every%20year",
```
jeffkaufman, line 50: This isn't a great source, and it's uncited.

I also think it's probably wrong? https://www.cdc.gov/chickenpox/vaccination-program-impact.html has "Before the [vaccination] program began, chickenpox used to be very common in the country. Each year in the early 90s, there were more than 4 million cases of chickenpox" ... "Now, chickenpox is rare in the United States, with fewer than 150,000 cases"

Maybe you can get the actual info from https://www.cdc.gov/chickenpox/health-departments/conducting-surveillance.html ?

pathogens/vzv.py
jeffkaufman, line 54: 335_000_000

pathogens/vzv.py
```diff
@@ -0,0 +1,119 @@
...
+    source="https://www.cdc.gov/shingles/hcp/clinical-overview.html#:~:text=An%20estimated%20one%20million%20cases%20of%20herpes%20zoster%20occur%20annually%20in%20the%20United%20States",
+)
+
+chickenpox_annual_infections = IncidenceAbsolute(
+    annual_infections=annual_new_VZV_infections.scalar,
+    tag="US-2023",
+    source="https://www.childrenshospital.org/conditions/chickenpox#:~:text=More%20than%2095%20percent%20of%20American%20adults%20have%20had%20chickenpox%20and%20about%204%2C000%2C000%20people%20get%20chickenpox%20every%20year",
+)
+
+shingles_duration = SheddingDuration(
```
simonleandergrimm, line 89: Add the CI in the document: three to five weeks.

pathogens/vzv.py
jeffkaufman, line 60: Don't do this sort of calculation manually.  Start with an `AboluteIncidence`.

pathogens/vzv.py
simonleandergrimm: https://www.cdc.gov/shingles/hcp/clinical-overview.html#:~:text=Last%20Reviewed%3A,Chickenpox%20Q%26As

pathogens/vzv.py
```diff
@@ -0,0 +1,119 @@
...
+    days=28,
+    source="https://www.nia.nih.gov/health/shingles#:~:text=Most%20cases%20of%20shingles%20last%20three%20to%20five%20weeks.",
+)
+
+chickenpox_duration = SheddingDuration(
+    days=12,
```
simonleandergrimm, line 95: Add CI: 10 to 14 days

simonleandergrimm (changes requested):

pathogens/vzv.py
```diff
@@ -0,0 +1,119 @@
...
+VZV_latent_prevalence = Prevalence(
+    infections_per_100k=population_above_34.infections_per_100k
+    + prevalence_from_postvaccine_cases.infections_per_100k,
+    active=Active.LATENT,
+)
+
+
+# Now, calculations for active chickenpox and shingles estimates
+
+shingles_annual_infections = IncidenceAbsolute(
```
jeffkaufman, line 77: Since the source has a relative incidence, probably better to use that?  Then we don't have to worry about whether we're using the same year's population.

pathogens/vzv.py
```diff
@@ -0,0 +1,119 @@
...
+    annual_infections=1000000,
+    tag="US-2023",
+    source="https://www.cdc.gov/shingles/hcp/clinical-overview.html#:~:text=An%20estimated%20one%20million%20cases%20of%20herpes%20zoster%20occur%20annually%20in%20the%20United%20States",
+)
+
+chickenpox_annual_infections = IncidenceAbsolute(
+    annual_infections=annual_new_VZV_infections.scalar,
+    tag="US-2023",
+    source="https://www.childrenshospital.org/conditions/chickenpox#:~:text=More%20than%2095%20percent%20of%20American%20adults%20have%20had%20chickenpox%20and%20about%204%2C000%2C000%20people%20get%20chickenpox%20every%20year",
```
jeffkaufman, line 86: As above, this is wrong

pathogens/vzv.py
```diff
@@ -0,0 +1,119 @@
...
+)
+
+shingles_duration = SheddingDuration(
+    days=28,
+    source="https://www.nia.nih.gov/health/shingles#:~:text=Most%20cases%20of%20shingles%20last%20three%20to%20five%20weeks.",
```
jeffkaufman, line 91: Let's record the range: `confidence_interval=[3,5]`

pathogens/vzv.py
```diff
@@ -0,0 +1,119 @@
...
+)
+
+chickenpox_duration = SheddingDuration(
+    days=12,
+    source="https://my.clevelandclinic.org/health/diseases/4017-chickenpox#:~:text=Chickenpox%20usually%20goes%20away%20after%2010%20to%2014%20days.",
```
jeffkaufman, line 96: here too: 10-14

pathogens/vzv.py
```diff
@@ -0,0 +1,119 @@
...
+)
+
+shingles_active_prevalence = Prevalence(
+    infections_per_100k=shingles_annual_infections.to_rate(US_Population)
+    .to_prevalence(shingles_duration)
+    .infections_per_100k,
+    active=Active.ACTIVE,
+)
```
jeffkaufman, line 104: ditto

pathogens/vzv.py
```diff
@@ -0,0 +1,119 @@
...
+
+chickenpox_active_prevalence = Prevalence(
+    infections_per_100k=chickenpox_annual_infections.to_rate(US_Population)
+    .to_prevalence(chickenpox_duration)
+    .infections_per_100k,
```
jeffkaufman, line 109: ditto

pathogens/vzv.py
```diff
@@ -0,0 +1,119 @@
...
+    active=Active.ACTIVE,
+)
+
+
+def estimate_prevalences():
+    return [
+        VZV_latent_prevalence,
+        chickenpox_active_prevalence,
+        shingles_active_prevalence,
```
jeffkaufman, line 118: I don't think this quite makes sense: active prevalence is chickenpox_active_prevalence + shingles_active_prevalence, no? [1]

The idea is, the amount of the virus we see in wastewater from active infections should be based on people actively shedding for both categories of infections.

pathogens/vzv.py
```diff
@@ -0,0 +1,119 @@
...
+    / US_Population.people
+    * 100000
+    * years_since_vaccine.scalar,
+    active=Active.LATENT,
+)
+
+# prevaccine cases estimate + postvaccine cases estimate
+VZV_latent_prevalence = Prevalence(
+    infections_per_100k=population_above_34.infections_per_100k
+    + prevalence_from_postvaccine_cases.infections_per_100k,
```
jeffkaufman, line 70: Don't pull infections_per_100k out of a `Prevalence` to make a new `Prevalence` -- that's very error prone.  Always use the built-in methods.  If you need to do something that can't be done with the built-ins consider adding one (in this case I think a new method would be very good) or post in `#p2ra-noisy` for guidance.

jeffkaufman: The method could look like:

```
def __add__(self: "Prevalence", other: "Prevalence") -> "Prevalence":
    assert self.active == other.active
    return Prevalence(
       infections_per_100k=self.infections_per_100k + other.infections_per_100k,
       inputs=[self, other],
       active=self.active,
    )
```

jeffkaufman (changes requested):
good start!

pathogens/vzv.py
jeffkaufman: Add go put in `__add__` now.

simonleandergrimm:
I will review this pull request as soon as we refocus on vzv, but I don't want it to be in the PR queue up until then.

PR #68, opened 2023-05-02
covid: import full Johns Hopkins dataset

jeffkaufman:
`southern-ca-cumulative-covid-cases.tsv` was an extract from `time_series_covid19_confirmed_US.csv` for just the three counties sampled in Rothman 2021.  Import the whole thing because there are a lot more counties we're interested in now.


simonleandergrimm (approved):

simonleandergrimm:
Looks good to me!

PR #71, opened 2023-05-03
Simon hbv

simonleandergrimm:
I used [2019 estimates for acute Hepatitis B](https://www.cdc.gov/hepatitis/statistics/2019surveillance/Introduction.htm#Technical:~:text=Reported%20cases%20of%20acute,with%20hospitalization%20information%20available) as the [2020 estimate was impacted by the pandemic](https://www.cdc.gov/hepatitis/statistics/2020surveillance/hepatitis-b.htm#:~:text=During%202020%2C%2044,19%20pandemic.), and later estimates aren't available. 

I added an issue to add spatially resolved Ohio data from 2021, found [here](https://odh.ohio.gov/wps/wcm/connect/gov/9c7b10d8-6d83-42e7-9f18-968c3a26963e/HBV+5+Year+Report+2021.pdf?MOD=AJPERES&CONVERT_TO=url&CACHEID=ROOTWORKSPACE.Z18_K9I401S01H7F40QBNJU3SO1F56-9c7b10d8-6d83-42e7-9f18-968c3a26963e-oqU9kQR?#page=3). Before adding more complexity through this, I'd be happy to get your thoughts on the existing estimate @jeffkaufman


pathogens/hbv.py
```diff
@@ -0,0 +1,89 @@
...
+    taxid=TaxID(10407),
+)
+
+
+hbv_us_2022_deaths_wonder = Number(
+    # Would need to find CFR to arrive at prevalence
+    number=1_240,
+    date="2022",
+    country="United States",
+    source="https://wonder.cdc.gov/controller/saved/D176/D341F095",
```
jeffkaufman, line 26: We shouldn't count on a WONDER saved search to persist indefinitely.  If you do keep this variable can you document here what settings you used?

pathogens/hbv.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+pathogen_chars = PathogenChars(
+    na_type=NAType.DNA,
+    enveloped=Enveloped.ENVELOPED,
+    taxid=TaxID(10407),
+)
+
+
+hbv_us_2022_deaths_wonder = Number(
```
jeffkaufman, line 21: It doesn't look like this is used anywhere in your estimate?

pathogens/hbv.py
```diff
@@ -0,0 +1,89 @@
...
+    source="https://wonder.cdc.gov/controller/saved/D176/D341F095",
+)
+
+dna_present_in_serum = SheddingDuration(
+    days=2 * 30.5,  # 2 months
+    date="1991",
+    source="https://pubmed.ncbi.nlm.nih.gov/1909458/",  # paywalled, data for
+    # this variable was read from figure 2, showing duration of presence of
+    # HBV-DNA in blood.
+)
```
jeffkaufman, line 35: This figure? 

image

Is this saying 2m duration, following a 2m incubation period?

It's suspicious that they show smooth curves crossing an axis break...

pathogens/hbv.py
```diff
@@ -0,0 +1,89 @@
...
+
+cdc_estimated_acute_2019 = IncidenceAbsolute(
+    annual_infections=20_700,
+    confidence_interval=(11_800, 50_800),  # 95% Bootstrap Confidence Interval
+    coverage_probability=0.95,
+    country="United States",
+    date="2019",
+    tag="us_2019",
+    source="https://www.cdc.gov/hepatitis/statistics/2019surveillance/Introduction.htm#Technical:~:text=20%2C700%20estimated%20infections%20(95%25%20CI%3A%2011%2C800%E2%80%9350%2C800)",
```
jeffkaufman, line 44: Add a comment like: `"During 2019, a total of 3,192 acute hepatitis B cases were reported to CDC, resulting in 20,700 estimated infections (95% CI: 11,800–50,800) after adjusting for case underascertainment and underreporting"`

pathogens/hbv.py
```diff
@@ -0,0 +1,89 @@
...
+    # is because the study takes into account the high burden of HBV in
+    # foreign-born individuals, and high-risk populations, both of which are
+    # likely underrepresented in NHANES.
+    infections=1_590_000,
+    confidence_interval=(1_250_000, 2_490_000),
+    active=Active.LATENT,
+    country="United States",
+    date="2020",
+    tag="us_2020",
+    source="https://journals.lww.com/ajg/fulltext/2020/09000/prevalence_of_chronic_hepatitis_b_virus_infection.20.aspx",
```
jeffkaufman, line 59: The paper has "It is likely that this range represents an underestimation of HBV prevalence because of the noted methodologic limitations. As such, the panel views its estimated US prevalence of chronic HBV to represent a conservative figure which will require modification as further evidence emerges."

We should try to adjust for this, right?

jeffkaufman (changes requested):
Looks good!

pathogens/hbv.py
simonleandergrimm: Yes it's not.  After a quick search I didn't fit a case-fatality rate (CFR) for HBV, which made me not use this number any further. 

Should I drop it or do another estimate, putting more time into finding a CFR?

jeffkaufman: I'd add a TODO like you have above for NHANES

jeffkaufman:
I don't see a response to my `The paper has "It is likely that this range represents an underestimation ..."` comment?

simonleandergrimm:
> I don't see a response to my `The paper has "It is likely that this range represents an underestimation ..."` comment?

I did answer, try this [link](https://github.com/naobservatory/p2ra/pull/71/files#r1183863572)

pathogens/hbv.py
jeffkaufman: Thanks, I was misreading it before.  The "HBV DNA" box was what I should have been looking at.

pathogens/hbv.py
jeffkaufman: I agree, this doesn't add up.  I did some playing with the numbers to see if maybe it would be clear where the difference came from, and their total (1,587,634) is suspiciously close to what you get if you include both "estimate" and "low" for us-born (303,327 + 137,314 + 1,075,069 + 72,014 = 1,587,724).

I think this is worth writing to the authors about to potentially get the paper corrected if it is indeed a mistake; want to draft something or should I?

Separately, I don't see any justification in the paper for why they're using the 'low' for foreign-born instead of 'estimate'?

What if we take their 1.7M and use that as is for now?

jeffkaufman (approved):

pathogens/hbv.py
simonleandergrimm: > Separately, I don't see any justification in the paper for why they're using the 'low' for foreign-born instead of 'estimate'?

Yeah I also don't see a justification for this.

> I think this is worth writing to the authors about to potentially get the paper corrected if it is indeed a mistake; want to draft something or should I?

 I'll draft something and send it to you for review! 

> What if we take their 1.7M and use that as is for now?

Would you still want to include the lower and upper estimate?

simonleandergrimm: Yes, I'll do that.

simonleandergrimm:
@jeffkaufman Made the final changes you requested. Shall I merge?

jeffkaufman (approved):

PR #74, opened 2023-05-03
Asher to prevalence

asherps:
Added a to_latent_prevalence method to convert from an IncidenceRate to a latent prevalence. Defined a YearsOfInfections object to make this possible. If we know the number of yearly infections and the approximate number of years these infections have been occurring for, we would then be able to convert the annual incidence to a latent prevalence.


.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+    "[python]": {
+        "editor.defaultFormatter": "ms-python.black-formatter"
+    },
+    "python.formatting.provider": "none"
+}
```
jeffkaufman, line 6: This doesn't belong in this PR

pathogen_properties.py
```diff
@@ -261,6 +266,15 @@ def to_prevalence(self, shedding_duration: SheddingDuration) -> Prevalence:
             active=Active.ACTIVE,
         )
 
+    def to_latent_prevalence(self, years: YearsOfInfections) -> Prevalence:
+        return Prevalence(
+            infections_per_100k=self.annual_infections_per_100k
+            * years.years
+            / 365,
```
jeffkaufman, line 273: Why this factor of 365?

jeffkaufman (changes requested):
Can you say more about how you intend to use this?  I'm nervous about this as a concept because it assumes (a) the incidence rate has been constant over the time period and (b) a constant population (no births, deaths, or migrants).

Happy to call (in person would be better but I'm wfh) if that would be easier?

simonleandergrimm:
I will review this pull request as soon as I/we refocus on vzv, but I don't want it to remain in the PR queue up until then.

PR #66, opened 2023-05-02
Asher rhinovirus

asherps:
Finished first draft of rhinovirus estimate

PR #78, opened 2023-05-04
pathogens: allow setting multiple taxids

jeffkaufman:
Prompted by https://github.com/naobservatory/p2ra/pull/66#discussion_r1185071121 and the same issue with Flu A vs B (which don't form a clade)

### DIFF
--- a/test.py
+++ b/test.py
@@ -111,18 +111,15 @@ class TestMGS(unittest.TestCase):
         sample_counts = mgs.load_sample_counts(self.repo)
         for p in ["sars_cov_2", "hiv"]:
             with self.subTest(pathogen=p):
-                self.assertIn(
-                    pathogens.pathogens[p].pathogen_chars.taxid,
-                    sample_counts,
-                )
+                for taxid in pathogens.pathogens[p].pathogen_chars.taxids:
+                    self.assertIn(taxid, sample_counts)
 
     def test_load_tax_tree(self):
         tree = mgs.load_tax_tree(self.repo)
         for p in ["sars_cov_2", "hiv", "norovirus"]:
             with self.subTest(pathogen=p):
-                self.assertIn(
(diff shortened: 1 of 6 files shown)


test.py
```diff
@@ -111,18 +111,15 @@ def test_load_sample_counts(self):
         sample_counts = mgs.load_sample_counts(self.repo)
         for p in ["sars_cov_2", "hiv"]:
```
dp-rice, line 112: Could you add a pathogen with more than one taxid to this list?

test.py
```diff
@@ -111,18 +111,15 @@ def test_load_sample_counts(self):
...
-                self.assertIn(
-                    pathogens.pathogens[p].pathogen_chars.taxid,
-                    sample_counts,
-                )
+                for taxid in pathogens.pathogens[p].pathogen_chars.taxids:
+                    self.assertIn(taxid, sample_counts)
 
     def test_load_tax_tree(self):
         tree = mgs.load_tax_tree(self.repo)
         for p in ["sars_cov_2", "hiv", "norovirus"]:
```
dp-rice, line 119: Same as above

dp-rice (approved):
As a change to the code, I don't have a problem with it. Scientifically, it would be nice to avoid doing this as much as possible. I feel a lot better about working with clades than with polyphyletic groups. I understand it might be necessary if the prevalence data for a virus systematically doesn't distinguish, though. I guess it's better to be able to represent this situation than not to.

I'd like to review again if and only if the suggested revised tests fail.

dp-rice:
@jeffkaufman forgot to reassign, sorry!

test.py
jeffkaufman: Fixed in #80 to use all pathogens.

test.py
jeffkaufman: I can once #66 goes in; there aren't any yet

PR #80, opened 2023-05-04
Hepatitis A: use the taxid the MGS Pipeline assigns reads to

jeffkaufman:
The code had [208726](https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Tree&id=208726&lvl=3&lin=f&keep=1&srchmode=1&unlock), which was more specific, but [Virus-Host DB](https://www.genome.jp/virushostdb/view/) doesn't know about 208726, only its parent [12092](https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Undef&id=12092&lvl=3&keep=1&srchmode=1&unlock).  Since we don't expect [Simian](https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Tree&id=12101&lvl=3&lin=f&keep=1&srchmode=1&unlock) or [Marsupial](https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=2200757&lvl=3&lin=f&keep=1&srchmode=1&unlock) HepA in the sequencing data I think this isn't a problem.

Prompted by trying to make the test more general for #78 and realizing it was failing on some of our pathogens.


simonleandergrimm (approved):
Thanks for spotting this!

jeffkaufman:
I'm guessing you meant to assign this back to me?

PR #75, opened 2023-05-03
Asher west nile virus

asherps:

PR #81, opened 2023-05-04
DS_Store: remove from repo, and add to gitignore

jeffkaufman:
Looks like this was accidentally added in #75


dp-rice:
@jeffkaufman, @asherps I just fixed this with 465e2c4. I guess this is why you use PRs instead of just pushing to `main`, haha.

jeffkaufman:
>  I guess this is why you use PRs instead of just pushing to main

For something this simple I'm ok with you pushing to main -- your change and mine were very nearly identical.

PR #86, opened 2023-05-09
Covid in Ohio: add county-specific estimates

jeffkaufman:
Now that we have county-level data for Spurbeck 2023, estimate covid on a per-county basis in Ohio.

Before:

```
$ ./summarize.py sars_cov_2 | grep Ohio | grep 2021-12-26
  8270.36 per 100k (Ohio, United States; 2021-12-26)
```

After:

```
$  ./summarize.py sars_cov_2 | grep Ohio | grep 2021-12-26
  9048.02 per 100k (Franklin, Ohio, United States; 2021-12-26)
  4338.53 per 100k (Greene, Ohio, United States; 2021-12-26)
  4878.58 per 100k (Lawrence, Ohio, United States; 2021-12-26)
  6463.61 per 100k (Licking, Ohio, United States; 2021-12-26)
  7128.86 per 100k (Lucas, Ohio, United States; 2021-12-26)
  5498.57 per 100k (Sandusky, Ohio, United States; 2021-12-26)
 13384.03 per 100k (Summit, Ohio, United States; 2021-12-26)
  7301.93 per 100k (Trumbull, Ohio, United States; 2021-12-26)
```

PR #92, opened 2023-05-11
Populations: import all county-level populations from census

jeffkaufman:
As suggested in https://github.com/naobservatory/p2ra/pull/86#pullrequestreview-1421270605


dp-rice (approved):
Looks good!

PR #93, opened 2023-05-11
Remove the need for tag= in most cases

jeffkaufman:
We added `tag=` to keep people from accidentally pairing an absolute count with the wrong population.  In cases where we're just checking that date and location match, though, we can do this automatically.

Keep `tag=` around for cases like hsv_2 where were talking about subpopulations and need to keep those clear.

### DIFF
--- a/test.py
+++ b/test.py
@@ -223,7 +223,6 @@ class TestPopulations(unittest.TestCase):
                 country="United States",
                 state="Rhode Island",
                 county="Bristol County",
-                tag="Bristol County, Rhode Island 2020",
             ),
         )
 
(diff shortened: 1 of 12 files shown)


simonleandergrimm (approved):

test.py
```diff
@@ -223,7 +223,6 @@ def test_county_state(self):
                 country="United States",
                 state="Rhode Island",
                 county="Bristol County",
-                tag="Bristol County, Rhode Island 2020",
```
simonleandergrimm, line 226: Not part of the review, just a question: Why do we need to look at this region specifically in `test.py`?

jeffkaufman: There isn't anything special about this region, we just needed to pick a region to make assertions about.

This is a common pattern in testing, where making assertions about every single region would be prohibitive, and making assertions about zero regions would risk missing things.

PR #96, opened 2023-05-11
target: remove some cases in which you need to `target()`

jeffkaufman:
If you're calling target to specify the location for an estimate, and the location you want is just the most specific location of any of your inputs, you don't need to call `target()` anymore.

### DIFF
--- a/pathogen_properties.py
+++ b/pathogen_properties.py
@@ -171,11 +171,23 @@ class Variable:
         if self.is_target and self._location():
             return self._location()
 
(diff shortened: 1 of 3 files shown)


simonleandergrimm (approved):

simonleandergrimm:
Should Dan also review this, or no?

jeffkaufman:
> Should Dan also review this, or no?

I don't think so?  I'll just give him a heads up about the API tweak

Issue #87, opened 2023-05-10
Improve estimate of prevalence from incidence


jeffkaufman:
Currently we convert incidence on a given day to prevalence on that day using the shedding duration (`S`).  But the current prevalence should depend on incidence over each of the last `S` days.

This is based on the idea that relative abundance is proportional to the amount of shedding that's going on.

jeffkaufman:
Which means I think we should conceive of the model as taking shedders per 100k as its main input

PR #97, opened 2023-05-12
Add date_source

jeffkaufman:
In working on fixing #87 I'm now doing something where incidence estimates targeting several different dates are being combined into a prevalence estimate targeting a single date. This exposed a bunch of issues with the way I'd initially set things up with `target()`. Now:

* When creating a `Variable` from other `Variable`s you can provide a `date_source`, as a way of saying which particular input variable controls the date the estimate is for.

* When you want to know when an estimate is for you just look at `parsed_start` and `parsed_end`, you don't have to investigate inputs.

* If you have an estimate that is targeting the wrong date, you can fix it with `dataclasses.replace(estimate, date_source=variable)`

`dataclasses.replace` can actually handle all of the cases where I was previously using `target()`, and is a much more robust way of handling this.

One downside is that `dataclasses.replace` skips `init=False` properties, so I needed to tweak a few properties to no longer be `init=False`.

### DIFF
--- a/pathogen_properties.py
+++ b/pathogen_properties.py
@@ -85,47 +85,73 @@ class Variable:
     date: InitVar[Optional[str]] = None
     start_date: InitVar[Optional[str]] = None
     end_date: InitVar[Optional[str]] = None
-    parsed_start: Optional[datetime.date] = field(init=False)
-    parsed_end: Optional[datetime.date] = field(init=False)
-    is_target: Optional[bool] = False
+    # In cases where an estimate is derived from multiple input variables with
+    # different dates, set date_source to the Variable that represents the date
+    # range this estimate is intended for.  For example, imagine we have:
+    #
+    #   population = Population(date="2020-04-01", ...)
+    #   prevalence = AbsolutePrevalence(date="2020-08-01", ...)
+    #   return prevalence.to_rate(population)
+    #
+    # There's no way for the consumer to know what date this estimate is for.
+    # So instead we do:
+    #
+    #   return prevalence.to_rate(population, date_source=prevalence)
+    #
+    # parsed_start / parsed_end are set from date_source if supplied, otherwise
+    # from start_date / end_date.
+    date_source: InitVar[Optional["Variable"]] = None
+    parsed_start: Optional[datetime.date] = None
+    parsed_end: Optional[datetime.date] = None
     taxid: Optional[TaxID] = None
     inputs: InitVar[Optional[Iterable["Variable"]]] = None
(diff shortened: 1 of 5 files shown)


pathogen_properties.py
```diff
@@ -85,47 +85,73 @@ class Variable:
...
+    #
+    # There's no way for the consumer to know what date this estimate is for.
+    # So instead we do:
+    #
+    #   return prevalence.to_rate(population, date_source=prevalence)
+    #
+    # parsed_start / parsed_end are set from date_source if supplied, otherwise
+    # from start_date / end_date.
+    date_source: InitVar[Optional["Variable"]] = None
+    parsed_start: Optional[datetime.date] = None
```
dp-rice, line 104: It's a bit of a shame that the type checker will always think `parsed_start` and `parsed_end` can be `None`

test.py
```diff
@@ -19,19 +19,16 @@ def test_hsv1_imported(self):
     def test_summarize_location(self):
         us_2019, la_2020 = pathogens.pathogens["hiv"].estimate_prevalences()
         self.assertEqual(us_2019.summarize_location(), "United States")
-        self.assertEqual(
-            us_2019.summarize_date(),
-            (datetime.date(2019, 1, 1), datetime.date(2019, 12, 31)),
-        )
+        self.assertEqual(us_2019.parsed_start, datetime.date(2019, 1, 1))
```
dp-rice, line 22: Nitpick: does this belong in `test_summarize_location`?

test.py
```diff
@@ -43,6 +40,13 @@ def test_properties_exist(self):
                 for estimate in pathogen.estimate_prevalences():
                     self.assertIsInstance(estimate, Prevalence)
 
+    def test_dates_set(self):
+        for pathogen_name, pathogen in pathogens.pathogens.items():
+            with self.subTest(pathogen=pathogen_name):
+                for estimate in pathogen.estimate_prevalences():
+                    self.assertTrue(estimate.parsed_start)
```
dp-rice, line 47: I think you want `self.assertIsNotNone`

dp-rice (approved):
Only minor comments. Sorry for the wait!

pathogen_properties.py
jeffkaufman: yeah

test.py
jeffkaufman: No it doesn't!  Fixed.

test.py
jeffkaufman: Neat!  That's new!

PR #98, opened 2023-05-12
incidence to prevalence: consider incidence over time

### LINKED ISSUE #87: Improve estimate of prevalence from incidence

jeffkaufman:
The number of currently infected people today is not today's incidence times the mean shedding duration.  It's much closer to the sum of incidences over the last shedding duration days.  Do that instead.

Here's how this changes our predictions for LA County.  Relatively subtle, but you can see it pushes peaks a bit into the future, and spreads them a bit horizontally, as you'd expect:

Screenshot 2023-05-12 at 2 44 53 PM

Fixes #87

### DIFF
--- a/pathogen_properties.py
+++ b/pathogen_properties.py
@@ -309,71 +309,63 @@ class SheddingDuration(Variable):
     # Incidences should contain daily estimates for target_day and at least the
     # prior `days` days.
     def prevalence_from_incidences(
-        self, target_day: datetime, incidences: list["IncidenceRate"]
+        self, target_day: datetime.date, incidences: list["IncidenceRate"]
     ) -> Prevalence:
         max_days = math.ceil(self.days)
         oldest_day = target_day - datetime.timedelta(days=max_days)
-        candiate_incidences: {}
+        candiate_incidences = {}
         for incidence in incidences:
+            assert incidence.parsed_start
+            assert incidence.parsed_end
             # Only handles daily incidences.
-            start_end = incidence.summarize_date()
-            assert start_end
-            start, end = start_end
-            assert start == end
-            incidence_day = start
-            if incidence_day > target_day or incidence_day < oldest_day:
+            assert incidence.parsed_start == incidence.parsed_end
+            incidence_day = incidence.parsed_start
+            if incidence_day > target_day or incidence_day <= oldest_day:
                 continue
 
             candiate_incidences[incidence_day] = incidence
 
         assert len(candiate_incidences) == max_days
 
-        infections_per_100k = 0
-        for offset in range(max_days):
-            incidence_day = target_day - datetime.timedelta(days=offset)
-            weight = 1
-            if offset == max_days - 1 and max_days != math.floor(self.days):
-                weight = self.days - math.floor(self.days)
-                infections_per_100k += (
-                    weight
-                    * candiate_incidences[
-                        incidence_day
-                    ].annual_infections_per_100k
-                    / 365
-                )
+        incidence_day = target_day
+        infections_per_100k = 0.0
+        n = 0
+        while incidence_day > oldest_day:
+            n += 1
+            weight = 1.0
+            if n > math.floor(self.days):
+                weight = math.floor(self.days) - n
+            infections_per_100k += (
+                weight
+                * candiate_incidences[incidence_day].annual_infections_per_100k
+                / 365
+            )
+            incidence_day -= datetime.timedelta(days=1)
 
         return Prevalence(
             infections_per_100k=infections_per_100k,
-            inputs=[self, shedding_duration],
+            inputs=[self, *candiate_incidences.values()],
             active=Active.ACTIVE,
-        ).target(date=target_day.isoformat())
+            date_source=candiate_incidences[target_day],
+        )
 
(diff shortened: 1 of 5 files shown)


pathogen_properties.py
```diff
@@ -301,6 +302,72 @@ def to_rate(self, population: Population) -> Prevalence:
...
+        max_days = math.ceil(self.days)
+        oldest_day = target_day - datetime.timedelta(days=max_days)
+        candiate_incidences = {}
+        for incidence in incidences:
+            assert incidence.parsed_start
+            assert incidence.parsed_end
+            # Only handles daily incidences.
+            assert incidence.parsed_start == incidence.parsed_end
+            incidence_day = incidence.parsed_start
+            if incidence_day > target_day or incidence_day <= oldest_day:
```
dp-rice, line 323: Minor: suggest streamlining this by inverting the condition and removing the `continue` since you're at the end of the block anyway.

pathogen_properties.py
```diff
@@ -301,6 +302,72 @@ def to_rate(self, population: Population) -> Prevalence:
...
+    # Incidences should contain daily estimates for target_day and at least the
+    # prior `days` days.
+    def prevalence_from_incidences(
+        self, target_day: datetime.date, incidences: list["IncidenceRate"]
+    ) -> Prevalence:
+        max_days = math.ceil(self.days)
+        oldest_day = target_day - datetime.timedelta(days=max_days)
+        candiate_incidences = {}
+        for incidence in incidences:
+            assert incidence.parsed_start
```
dp-rice, line 318: An idea: It's kind of annoying that `parsed_start` and `parsed_end` are now `Optional` so that you have to assert these everywhere you use them. Consider making a pair of getter methods that do the assertion and return `date`.

pathogen_properties.py
```diff
@@ -301,6 +302,72 @@ def to_rate(self, population: Population) -> Prevalence:
...
+            assert incidence.parsed_end
+            # Only handles daily incidences.
+            assert incidence.parsed_start == incidence.parsed_end
```
dp-rice, line 321: Similarly, it might be nice to hide this logic behind an `is_daily(self) -> bool` method.

pathogen_properties.py
```diff
@@ -301,6 +302,72 @@ def to_rate(self, population: Population) -> Prevalence:
...
+            incidence_day = incidence.parsed_start
+            if incidence_day > target_day or incidence_day <= oldest_day:
+                continue
+
+            candiate_incidences[incidence_day] = incidence
+
+        assert len(candiate_incidences) == max_days
+
+        incidence_day = target_day
```
dp-rice, line 330: I think this could be made clearer and more general by framing the problem as approximating the integral of the incidence over the shedding period using a fixed set of (time, incidence) pairs. You could then replace this section with a call to `scipy.integrate.simpson` (https://docs.scipy.org/doc/scipy/tutorial/integrate.html#integrating-using-samples).
Note that I'm already adding `scipy` as a dependency in the branch I'm working on, so it won't add any new dependencies.

pathogen_properties.py
```diff
@@ -301,6 +302,72 @@ def to_rate(self, population: Population) -> Prevalence:
...
+        self, incidences: list["IncidenceRate"]
+    ) -> list[Prevalence]:
+        dates = set()
+        for incidence in incidences:
+            assert incidence.parsed_start
+            dates.add(incidence.parsed_start)
+        first_date = min(dates) + datetime.timedelta(days=math.ceil(self.days))
+        last_date = max(dates)
+
+        prevalences = []
```
dp-rice, line 362: Any reason not to make this a list comprehension?

pathogen_properties.py
```diff
@@ -301,6 +302,72 @@ def to_rate(self, population: Population) -> Prevalence:
...
+            infections_per_100k=infections_per_100k,
+            inputs=[self, *candiate_incidences.values()],
+            active=Active.ACTIVE,
+            date_source=candiate_incidences[target_day],
+        )
+
+    def prevalences_from_incidences(
+        self, incidences: list["IncidenceRate"]
+    ) -> list[Prevalence]:
+        dates = set()
```
dp-rice, line 355: If you take my `get_start` method suggestion, this can be a set comprehension.

dp-rice (changes requested):
Seems like the right idea even if we end up changing our definition of "prevalence" and want to adjust it. My main suggestion is the one about recasting the problem as numerical integration.

pathogen_properties.py
jeffkaufman: That seems to make it a lot more complex, and also seems like it wouldn't be the right model in a future where we model non-uniform shedding?

dp-rice (approved):
Looks good. I still think we'll want to switch the calculation to a standard numerical integration method eventually. That will make it simpler to accommodate variable shedding and incidence estimates that are unevenly spaced.

dp-rice:
Sorry, I always forget to reassign!

PR #99, opened 2023-05-17
Clean up location comparison

dp-rice:
- Pull updated county labels for MGS samples and use them correctly
- Use `target_location` to get the locations for prevalence estimates so that it correctly looks down the chain of estimates rather than just at the tip.
- Package the code that looks up covid prevalence for each MGS sample into a function and add a unit test so we know if this breaks.


dp-rice:
Note this is rebased onto #98

jeffkaufman (approved):

PR #94, opened 2023-05-11
Asher ebv denmark

asherps:
Estimate for EBV's prevalence in Denmark using a 1983 EBV seroprevalence study from Denmark.

PR #101, opened 2023-05-18
Add location_source

jeffkaufman:
We need this because in #94 we're going to be generating an estimate for one place based on data from a different specific place, and the system doesn't know which location we're targeting.

### DIFF
--- a/pathogen_properties.py
+++ b/pathogen_properties.py
@@ -102,6 +102,9 @@ class Variable:
     # parsed_start / parsed_end are set from date_source if supplied, otherwise
     # from start_date / end_date.
     date_source: InitVar[Optional["Variable"]] = None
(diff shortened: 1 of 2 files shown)


dp-rice:
@jeffkaufman does this make `target_location` obsolete?

jeffkaufman:
@dp-rice `target_location` is still doing something in cases where `location_source` is not set (and it's often not set) so no. But I could move that logic to `__post_init__` and remove `target_location`?

dp-rice:
From my perspective all that matters is that I have a single clear interface for getting locations that works however the variable was constructed. Whatever you think makes the most sense. Maybe it would be good to rename it to `get_location` to be parallel to `get_date`(s)

asherps (approved):

PR #102, opened 2023-05-18
Asher hbv

asherps:
Created an ohio-specific acute hbv estimate for 2021.


pathogen_properties.py
```diff
@@ -207,7 +211,7 @@ def target_location(
         countries = set(i.country for i in inputs if i.country)
         states = set(i.state for i in inputs if i.state)
         counties = set(i.county for i in inputs if i.county)
-
+        print(states)
```
jeffkaufman, line 214: rm

pathogen_properties.py
```diff
@@ -255,6 +270,9 @@ class Population(Taggable):
 
     people: float
 
+    def __sub__(self, other: "Population") -> "Population":
+        return dataclasses.replace(self, people=self.people - other.people)
```
jeffkaufman, line 274: `other` needs to end up in `inputs`

pathogens/rhinovirus.py
```diff
@@ -0,0 +1,264 @@
+from pathogen_properties import *
```
jeffkaufman, line 1: Your changes to rhinovirus.py don't belong in this PR.

In general, when you start a new PR that should look like:

```
git checkout main
git branch -b asher-some-feature
...
```

I suspect you skipped the `git checkout main` here?

jeffkaufman: Same with your `pathogen_properties.py` changes.  I think this PR should only include `hbv.py`?

jeffkaufman (changes requested):

pathogens/hav.py
jeffkaufman, line 1: Why are you deleting `hav.py`?  This PR shouldn't affect `hav.py` either way.

pathogens/hbv.py
```diff
@@ -66,9 +66,25 @@
 )
 
 
+ohio_acute_incidence_2021 = IncidenceRate(
+    # This source reports both an acute and a chronic incidence. I think it
+    # makes sense to add these 2 numbers for an acute estimate, since chronic
+    # cases still have an acute phase in which they shed.
```
jeffkaufman, line 72: @simonleandergrimm does this seem reasonable to you?  (Looks fine to me)

jeffkaufman (changes requested):

pathogens/hbv.py
asherps: I think if this seems reasonable to you, I think it's good to go? I don't think there are any other changes I need to make (there was no reason to remove hav, as jeff noted above, but it shouldnt matter either way if this is merged into main)

simonleandergrimm: This isn't needed. If you look at the table more specifically, the table gives **acute** and **total** numbers. The authors then say that
```
Total hepatitis equals all hepatitis B cases,both "acute" and "chronic".
```
So just using total is the way to go (which Asher is already doing). I told @asherps to change the wording of the comment to not make it sound like there is manual addition.

pathogens/hbv.py
```diff
@@ -66,9 +66,30 @@
...
 
+ohio_total_incidence_2021 = IncidenceRate(
+    # This source reports both an acute and a total incidence. The total
+    # incidence is the sum of the acute and chronic incidences. I think it
+    # makes sense to report the total incidence, since chronic
+    # cases still have an acute phase in which they shed. The number of
+    # chronic cases has some lag attached, since chronic cases are probably
+    # not identified during their acute phase. However, the reported chronic
+    # incidence in any given year should still be around the true number of
+    # new chronic cases for that year.
```
simonleandergrimm, line 77: What do you mean by true number of new chronic cases here?

simonleandergrimm: I would also shorten this paragraph to one or two sentences, it's a bit wordy atm.

simonleandergrimm (changes requested):

pathogens/hbv.py
```diff
@@ -66,9 +66,29 @@
...
 
 
+ohio_total_incidence_2021 = IncidenceRate(
+    # This source reports both an acute and a total (acute+chronic) incidence.
+    # I think it makes sense to report the total incidence, since chronic
+    # cases still have an acute phase in which they shed. The number of
+    # chronic cases in any given year has some lag attached, since chronic
+    # cases are probably not identified during their acute phase. However, the
+    # reported chronic incidence in any given year should still be
+    # approximately the number of new chronic cases for that year.
```
simonleandergrimm, line 76: We talked about this yesterday, but I am unsure what you are arguing for or against here, could you explain?

```
However, the reported chronic incidence in any given year should still be approximately the number of new chronic cases for that year.
```
What do you mean by this? I would think reported incidence very likely underestimates the true number of true cases.

simonleandergrimm (commented):
The estimate looks good, I just have a question on your comment on line 70.

pathogens/hbv.py
asherps: I'm arguing that chronic cases should be counted because they're our best guess at true chronic incidence for a current year, even some of the new chronic cases identified are from years ago. I updated the comment again, see what you think

pathogens/hbv.py
```diff
@@ -66,9 +66,29 @@
...
 
 
+ohio_total_incidence_2021 = IncidenceRate(
+    # This source reports both an acute and a total (acute+chronic) incidence.
+    # I think it makes sense to report the total incidence, since chronic
+    # cases still have an acute phase in which they shed. The number of
+    # chronic cases in any given year has some lag attached, since chronic
+    # cases are probably not identified during their acute phase, but this
+    # number still roughly represents our most recent guess at new chronic
+    # cases for 2021
```
simonleandergrimm, line 76: Added it to my agenda to talk through this tomorrow

simonleandergrimm:
Superseded by #122 . I will thus close this PR.

PR #104, opened 2023-05-19
location: work like date

### LINKED ISSUE #103: Add get_location that works like get_dates
Our location handling is silly, and should work like our dates

jeffkaufman:
Instead of resolving location conflicts when someone asks an estimate for its location, resolve them at the time the estimate is created.  This makes location work like date, and reduces the chances that we have broken estimates we haven't noticed.

Fixes #103


dp-rice (approved):
Looks good!

PR #108, opened 2023-05-22
Stop converting incidences to prevalences

jeffkaufman:
We've decided that in cases where the public health data gives us incidence we'll feed incidence into the model, and not try to convert to a prevalence first.  See https://docs.google.com/document/d/1qZ0mnqD4NgesSwp5hU2Nc5EXFkSSJkG3GQlPbtog5Wg/edit for full reasoning.

Example `summarize.py` output:
```
$ ./summarize.py hbv
hbv
   6.31 per 100k/y (United States; 2019)
   519.15 per 100k (United States; 2020)
```

### DIFF
--- a/fit_covid.py
+++ b/fit_covid.py
@@ -10,11 +10,11 @@ from mgs import BioProject, Enrichment, MGSData, Sample, SampleAttributes
 from pathogens import pathogens
 
 
-def prevalence_by_state_county_date(
+def incidence_by_state_county_date(
     pathogen: str,
 ) -> dict[tuple[str, str, date], float]:
     prevs = {}
-    for estimate in pathogens[pathogen].estimate_prevalences():
+    for estimate in pathogens[pathogen].estimate_incidences():
         country, state, county = estimate.get_location()
         assert country == "United States"
         assert state is not None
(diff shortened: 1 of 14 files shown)


fit_covid.py
```diff
@@ -10,26 +10,26 @@
 from pathogens import pathogens
```
dp-rice, line 10: I think some of these changes won't work with the functions in `stats.py` that are called, but I'm working on a PR that substantially rewrites both files, so don't worry about it.

dp-rice: (Also as part of that PR I'll write some more tests so that we catch this sort of thing.)

summarize.py
```diff
@@ -17,46 +44,28 @@ def start(pathogen_names):
...
-                prevalence.rjust(18),
+                details.rjust(18),
                 location,
                 date,
                 taxid,
             )
-            to_print.append(((location, str(date)), line))
+            to_print.append(((location, date), line))
+
+        for estimate in pathogen.estimate_prevalences():
```
dp-rice, line 61: Sorry if I'm missing something obvious, but does this print whether the estimate is an incidence or a prevalence? If not, it probably should.

dp-rice (approved):
Looks good to me. Only minor comments.

PR #125, opened 2023-05-31
Compare prevalence/incidence estimates with MGS data

dp-rice:
Not sure whether this should be merged into main, but could be useful for discussion:

Added `data_comparison.py`, which generates `estimate_table.tsv`, which says whether we have an incidence/prevalence estimate for each RNA virus for each MGS sampling time/location in Rothman, Crits-Christoph, and Spurbeck.

### DIFF
--- a/data_comparison.py
+++ b/data_comparison.py
@@ -1,0 +1,72 @@
+#!/usr/bin/env python3
+from pathlib import Path
+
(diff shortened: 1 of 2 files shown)


jeffkaufman:
Superseded by #133

PR #135, opened 2023-06-02
Fix "Montogmery County" typo

dp-rice:
Bump MGS data tag and fix `target_counties`.

Should now match for all `sars_cov_2`:
```
$  ./check-matches.py | grep sars_cov_2

```


jeffkaufman (approved):

jeffkaufman:
Going ahead with merging this without waiting for @dp-rice because it's so simple and my branch depends on it

PR #140, opened 2023-06-03
Simon norovirus | added comment

simonleandergrimm:


pathogens/norovirus.py
```diff
@@ -91,7 +91,7 @@ def load_nors_outbreaks() -> (
     # Click "Download all NORS Dashboard data (Excel)."
     # Exported from Google Sheets as CSV.
     #
-    # Data runs through the end of 2021.
+    # Data entries start in 1971 and run through the end of 2021.
```
jeffkaufman, line 94: Maybe something about how we're only using from HISTORY_START (2012) onward?

jeffkaufman (approved):

PR #145, opened 2023-06-06
Divide predictions by taxids

jeffkaufman:
When you call pathogen.estimate_incidences or pathogen.estimate_prevalences you can get estimates for a range of taxids.  For example, with Norovirus you can get ones for (a) Norovirus overall, (b) Group I, and (c) Group II.

To let modeling handle each group of taxids separately, add a converter method, by_taxids, that takes a list of estimates and breaks it down by which taxids they're for.

I've updated the stats.build_model interface for this, and test_build_model and fit.py accordingly, but I haven't verified that my changes to fit.py actually work.

Separately, I ran into an issue with stats.py where a long import line was getting formatted differently between black and isort, and needed to tell isort to use black's rules.

### DIFF
--- a/stats.py
+++ b/stats.py
@@ -12,14 +12,7 @@ import stan  # type: ignore
 from scipy.stats import gamma, norm  # type: ignore
 
 from mgs import BioProject, Enrichment, MGSData, Sample, SampleAttributes
-from pathogen_properties import (
-    PathogenChars,
-    Predictor,
-    TaxID,
-    Variable,
-    by_taxids,
-)
-from pathogens import pathogens
+from pathogen_properties import Predictor, TaxID, Variable
 
 county_neighbors = {
     "Los Angeles County": [
@@ -104,7 +97,7 @@ def lookup_variables(
 
     if not qualities:
         return []
-    best_quality = max(quality for (quality, var) in qualities)
+    best_quality = max(quality for (quality, _) in qualities)
     return [var for (quality, var) in qualities if quality == best_quality]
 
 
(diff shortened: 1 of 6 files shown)


stats.py
```diff
@@ -231,41 +237,32 @@ def plot_posterior_samples(
...
-        predictors = pathogen.estimate_prevalences()
-    else:
-        raise ValueError(
-            f"{predictor} must be one of 'incidence' or 'prevalence'"
-        )
     data = [
         DataPoint(
             sample=s,
             attrs=attrs,
             viral_reads=mgs_data.viral_reads(bioproject, taxids)[s],
```
dp-rice, line 258: It just occurred to me that my `mgs_data.viral_reads` assumes that the TaxIDs are not nested under one another, otherwise they will double count. We probably don't want that, do we? Do we ever give it nested TaxIDs here?

stats.py
```diff
@@ -231,41 +237,32 @@ def plot_posterior_samples(
...
-            predictor=choose_predictor(lookup_variables(attrs, predictors)),
+            predictor=choose_predictor(
+                lookup_variables(attrs, grouped_predictors)
+            ),
         )
+        for taxids, grouped_predictors in by_taxids(
```
dp-rice, line 263: Isn't this going to make a big list of datapoints with all the taxids mixed together? We should be building a separate model for each pathogen that we want to model. (For now. Eventually, we'll want to consider them together, but that's not how the model currently works)

dp-rice (commented):
The part that generates the predictors looks good, but I don't think it's interacting correctly with the model. If it sounds good to you, I'll make a PR into this one with some changes to `build_model` and `fit.py`

dp-rice (approved):
Just made a PR with some changes. If those make sense to you, then this can go in.

dp-rice:
Actually, just realized one thing: `TestPathogensMatchStudies` should check that there's data for each set of taxids, not just for each pathogen.

jeffkaufman:
> Actually, just realized one thing: TestPathogensMatchStudies should check that there's data for each set of taxids, not just for each pathogen.

Done in b0e0bad6b59ec756e51030293264ac8bad18f6a4

PR #150, opened 2023-06-09
Simon hcv

simonleandergrimm:
As discussed, I dropped chronic incidence estimates for Hep C, as this is hard to work with.

### DIFF
--- a/pathogens/hcv.py
+++ b/pathogens/hcv.py
@@ -99,8 +99,11 @@ acute_underreporting_factor = Scalar(
 
 ohio_counties_case_rates = {
     # source: https://odh.ohio.gov/wps/wcm/connect/gov/ec0dec22-1eea-4d17-a86a-ac4bc35be4d3/HCV+5+Year+Report+2021.pdf?MOD=AJPERES&CONVERT_TO=url&CACHEID=ROOTWORKSPACE.Z18_K9I401S01H7F40QBNJU3SO1F56-ec0dec22-1eea-4d17-a86a-ac4bc35be4d3-oqU9kQ8
-    # We ended up not using the total information listed here, as chronic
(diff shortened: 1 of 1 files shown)


pathogens/hcv.py
```diff
@@ -121,6 +99,11 @@
 
 ohio_counties_case_rates = {
     # source: https://odh.ohio.gov/wps/wcm/connect/gov/ec0dec22-1eea-4d17-a86a-ac4bc35be4d3/HCV+5+Year+Report+2021.pdf?MOD=AJPERES&CONVERT_TO=url&CACHEID=ROOTWORKSPACE.Z18_K9I401S01H7F40QBNJU3SO1F56-ec0dec22-1eea-4d17-a86a-ac4bc35be4d3-oqU9kQ8
+    # We ended up not using total cases, which could be used to arrive at
+    # chronic incidence by subtracting acute cases [total - acute]. This is
+    # because because we i) do not have an underreporting factor for chronic
+    # incidence, and ii) prevalence is a better measure for the chronic
+    # version of Hep C  prevalence
```
jeffkaufman, line 106: I thought the reason was that chronic cases start off as (non-reported) acute cases, and so we're effectively already counting these when we look at underreporting-adjusted acute cases?

jeffkaufman (approved):

pathogens/hcv.py
simonleandergrimm: Hm, not really. Hepatitis C always starts as acute and gets only defined as chronic if HCV RNA is detectable over a 6 months period.

Hepatitis C chronic incidence (i.e. detection) lags acute incidence. It will thus not correspond to current acute incidence if acute incidence was lower or higher in the past.

PR #151, opened 2023-06-09
Don't emit zeros for any predictor

### LINKED ISSUE #148: Some viruses give 0 incidences, which break the model
The model is in log space, and 0 is very bad.

Check to make sure there are no zeros in matched data.  When we find them, bump them up based on the granularity.  For now, figure 1/10 of if we had seen 1 in our calibration data, which is kind of unprincipled but better than crashing.  Dan will review.

jeffkaufman:
Zero is no good: we take the log of values, and if they're zero that explodes.

For Norovirus the issue was that we look at the I/II ratio and if either of those is zero then we'll emit a zero for that pathogen; in that case we can fall back to the annual average ratio.

For Influenza the issue was that some weeks had zero positive tests; in that case we somewhat unprincipledly treat it as if there were really 0.1 positive tests.

Fixes #148, with a test to verify it stays fixed.

### DIFF
--- a/pathogen_properties.py
+++ b/pathogen_properties.py
@@ -511,3 +511,14 @@ def by_taxids(
 
         out[taxids].append(predictor)
     return out
+
+
+# We don't want to predict zero of any pathogen, both because they almost never
+# go truely to zero and because modeling will be taking logs of our output.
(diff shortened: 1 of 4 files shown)


dp-rice (approved):
Looks good, thanks!

PR #152, opened 2023-06-09
Added a test to identify duplicate prevalences/incidences. Dropped duplicate HSV-1 and HSV-2 seroprevalence estimates.

simonleandergrimm:
We sometimes had multiple prevalence/incidence estimates for the same pathogen, time, and place. This can lead to errors in downstream analyses, and we instead want to either combine estimates, or drop estimates, with the goal of having only one estimate per time, place, and pathogen.

This PR adds a test, separately spotting duplicate prevalence and incidence estimates. It also drops duplicate estimates for HSV-1 and HSV-2, which we identified through this test.

### DIFF
--- a/pathogen_properties.py
+++ b/pathogen_properties.py
@@ -39,11 +39,6 @@ class Active(Enum):
     LATENT = "Latent"
 
 
-class Primary(Enum):
-    PRIMARY = "Primary"
-    SECONDARY = "Secondary"
-
-
 TaxID = NewType("TaxID", int)
 
 
@@ -330,7 +325,6 @@ class Prevalence(Predictor):
     """What fraction of people have this pathogen at some moment"""
 
     active: Active
-    primary: Optional[Primary] = None
     infections_per_100k: float
 
     def get_data(self) -> float:
@@ -343,7 +337,6 @@ class Prevalence(Predictor):
             active=self.active,
             date_source=self,
             location_source=self,
-            primary=self.primary,
         )
 
     def __truediv__(self, scalar: Scalar) -> "Prevalence":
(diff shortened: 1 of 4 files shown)


test.py
```diff
@@ -76,6 +76,36 @@ def test_by_taxids(self):
...
+                    (
+                        "prevalence",
+                        pathogen.estimate_prevalences(),
+                    ),
+                    (
+                        "incidence",
+                        pathogen.estimate_incidences(),
+                    ),
+                ]:
+                    seen = set()
```
jeffkaufman, line 92: `seen` should be inside the loop, and then you don't need to include `taxids` in `key`

pathogen_properties.py
```diff
@@ -39,6 +39,11 @@ class Active(Enum):
     LATENT = "Latent"
 
 
+class Primary(Enum):
```
jeffkaufman, line 42: What's the argument for adding a primary/secondary distinction instead of just only returning our best estimate for every location+timeperiod+taxid?

jeffkaufman (changes requested):

simonleandergrimm:
By returning, do you mean not adding them to `estimate_*`, while still
keeping them in the `pathogen.py` file? I'd rather not do that, as it might
be confusing.

Given our current set of pathogens, I can simply cut the HSV-1, and HSV-2
duplicate estimates, as those represent the raw NHANES data, while the CDC
estimates that are based on that data. I will instead mention the data in
the comments of the CDC-based prevalence estimates. In that case, I'd also
remove `class Primary(Enum):`

Does that sound good?

jeffkaufman:
@simonleandergrimm I think we should be generating a single best effort estimate, either by cutting low-quality estimates or combining multiple estimates.  Your choice which!

simonleandergrimm:
I cut the estimates, ready for re-review!

jeffkaufman (approved):
Looks good, but update the PR title and add a summary of what you're doing and why as the PR description

PR #154, opened 2023-06-12
Simon hep b c extension

simonleandergrimm:
- Switched to using the acute Hep C CDC estimate from 2019 instead of the 2020 estimate. The CDC underreporting factors in 2020 for Hepatitis B and C were the same as in 2019, which seems wrong given the pandemic should have impacted case reporting quite a bit.
- Extrapolated acute and chronic Hep B, and acute Hep C estimates to 2019-2021.

### DIFF
--- a/pathogens/hbv.py
+++ b/pathogens/hbv.py
@@ -41,42 +41,6 @@ cdc_acute_underreporting_factor_2020 = Scalar(
 )
 
 
-cdc_estimated_acute_2019 = IncidenceAbsolute(
-    annual_infections=20_700,
-    # During 2019, a total of 3,192 acute hepatitis B cases were reported to
-    # CDC, resulting in 20,700 estimated infections (95% CI: 11,800–50,800)
-    # after adjusting for case underascertainment and underreporting.
-    confidence_interval=(11_800, 50_800),  # 95% Bootstrap Confidence Interval
-    coverage_probability=0.95,
-    country="United States",
-    # I picked 2019 rates, as estimated rates in 2020 were 30% lower, even
-    # while deaths stayed the same, pointing to undercounting due to
-    # COVID-19: https://www.cdc.gov/hepatitis/statistics/2020surveillance/introduction/national-profile.htm#:~:text=hepatitis%20B%20transmission.-,Data%20from%20death%20certificates%20filed%20in%20the%20vital%20records%20offices%20of,same%20as%20the%20rate%20during%202019%20(0.42%20deaths%20per%20100%2C000%20population).,-Hepatitis%20C
-    date="2019",
-    source="https://www.cdc.gov/hepatitis/statistics/2019surveillance/Introduction.htm#Technical:~:text=20%2C700%20estimated%20infections%20(95%25%20CI%3A%2011%2C800%E2%80%9350%2C800)",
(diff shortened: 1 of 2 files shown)


pathogens/hbv.py
```diff
@@ -77,28 +88,37 @@
...
-    source="https://odh.ohio.gov/know-our-programs/viral-hepatitis/data-statistics/hbv_5yr_report",
-)
-
-
 def estimate_prevalences() -> list[Prevalence]:
+    chronic_2020 = estimated_chronic_us_2020.to_rate(us_population(year=2020))
+    # Hep B acute and chronic cases should be approximately constant, so we
+    # can use chronic 2020 estimates for 2019 and 2021, and acute 2019
+    # estimates for 2020 and 2021.
+    chronic_2019 = dataclasses.replace(
```
jeffkaufman, line 96: Why do we need a 2019 estimate?

pathogens/hbv.py
```diff
@@ -77,28 +88,37 @@
...
+        chronic_2019,
+        chronic_2020,
+        chronic_2021,
     ]
 
 
 def estimate_incidences() -> list[IncidenceRate]:
+    acute_2019 = estimated_acute_2019.to_rate(us_population_2019)
+
+    acute_2020 = dataclasses.replace(
```
jeffkaufman, line 113: comment here about why the extrapolation is ok

pathogens/hcv.py
```diff
@@ -170,7 +179,7 @@ def estimate_incidences():
 
 
 def estimate_prevalences() -> list[Prevalence]:
-    estimates = [
-        estimated_current_infection_us_2013_2016,
-    ]
+    # This estimate being from 2016, I (Simon) do not want to transfer it to
+    # 2019-202 without first doing further research.
+    estimates = [estimated_chronic_us_2013_2016]
```
jeffkaufman, line 184: just `return [estimated_chronic_us_2013_2016]`

jeffkaufman (changes requested):
looks good!

pathogens/hcv.py
jeffkaufman: Actually, we may want to return `[]` here since `estimated_chronic_us_2013_2016` won't match any of our mgs samples

pathogens/hbv.py
simonleandergrimm: We do not, I removed it!

pathogens/hcv.py
simonleandergrimm: I looked into the acute case numbers of Hep C (https://www.cdc.gov/hepatitis/statistics/2020surveillance/hepatitis-c/figure-3.1.htm), and they have ticked up slightly in the past few years. Mortality, on the other hand, has dropped slightly (https://www.cdc.gov/hepatitis/statistics/2020surveillance/hepatitis-c/table-3.7.htm). On balance, I think it's unlikely that the chronic rate dropped or increased from 1% in 2016 by more than 0.2 percentage points.

I thus extrapolated the estimate after all.

pathogens/hbv.py
jeffkaufman: Also say something about why you wouldn't expect the covid 19 pandemic to affect hbv incidence (or why you do think it probably affected it but you're extrapolating anyway)

PR #162, opened 2023-06-16
Add stubs for BKV, JCV, MCV, and AAV2

jeffkaufman:
We are planning to add estimates for these four viruses (https://twist.com/a/197793/ch/661244/t/4455246/) and it would be helpful to test them in modeling at this stage.  We expect their population prevalence to be constant over time, which means modeling will only be off by a (very important eventually but not important right now) constant factor.

### DIFF
--- a/pathogens/aav2.py
+++ b/pathogens/aav2.py
@@ -1,0 +1,31 @@
+import dataclasses
+
+from pathogen_properties import *
+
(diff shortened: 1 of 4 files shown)


dp-rice (approved):

dp-rice:
@jeffkaufman, I just merged this so I can use it now.

jeffkaufman:
Great!

PR #163, opened 2023-06-16
Simon jc virus

simonleandergrimm:
Adds a new estimate for JC Virus, one of three polyomaviruses that we want to cover due to their high prevalence and their presence in RNA sequencing data.

Next week, I will improve this estimate by adjusting US data by age cohort size (#164).

### DIFF
--- a/pathogens/jcv.py
+++ b/pathogens/jcv.py
@@ -1,4 +1,3 @@
-
 from pathogen_properties import *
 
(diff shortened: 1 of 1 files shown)


pathogens/jcv.py
```diff
@@ -12,18 +10,74 @@
 )
 
 
+ch_2009_seroprevalence = Prevalence(
+    infections_per_100k=0.58 * 100_000,
+    # "JCV IgG seroprevalence of 58% (231 of 400)"
```
jeffkaufman, line 15: Flag here that this was blood donors too

jeffkaufman (approved):

PR #170, opened 2023-06-21
Simon aav update

simonleandergrimm:
Adds AAV5 and AAV6, and updated seroprevalence of AAV2 to represent a the unweighted global seroprevalence average.

@jeffkaufman Let me know if one PR for AAV 5 and 6 is fine or if I should have separated these.


pathogens/aav5.py
```diff
@@ -0,0 +1,61 @@
...
+pathogen_chars = PathogenChars(
+    na_type=NAType.DNA,
+    enveloped=Enveloped.NON_ENVELOPED,
+    taxid=TaxID(82300),
+)
+
+seroprevalence_hemophilia_global_2021 = Prevalence(
+    infections_per_100k=0.297 * 100_000,
+    # "Factoring in the prevalence of HA in the countries being assayed, the
+    # global weighted average of AAV5 seroprevalence in people with HA was 29.
```
jeffkaufman, line 20: nit: wrapping on a decimal point is hard to distinguish from end-of-sentence

jeffkaufman (approved):

pathogens/aav5.py
```diff
@@ -0,0 +1,61 @@
...
+
+pathogen_chars = PathogenChars(
+    na_type=NAType.DNA,
+    enveloped=Enveloped.NON_ENVELOPED,
+    taxid=TaxID(82300),
+)
+
+seroprevalence_hemophilia_global_2021 = Prevalence(
+    infections_per_100k=0.297 * 100_000,
+    # "Factoring in the prevalence of HA in the countries being assayed, the
```
jeffkaufman, line 19: > Factoring in the prevalence of HA in the countries being assayed

That's a weird thing to do for our purposes.  Why not use the 34.8 global non-weighted average?

jeffkaufman (changes requested):
Since this source also has AAV8 and AAVrh10, why not include them too?

simonleandergrimm:
@jeffkaufman Flagging that this is still open.

jeffkaufman (approved):

PR #171, opened 2023-06-21
Hierarchical model checking

### LINKED ISSUE #169: Assess the sensitivity to the prior on mu
Currently the prior on mu is constraining the posterior on the low end, and the posterior is not peaked at zero. Try wider (possibly flat) priors on mu and see how that changes the results.

dp-rice:
Resolves #169. Made some more plots to assess the model fit. I found that:

* The prior on `mu` (the centered log P2RA coefficient) was too narrow (std=2), so that it was constraining the posterior even for viruses where we had plenty of data (e.g., Sars-CoV-2).
* If I made it too wide (std=10), the sampler had trouble mixing. The posterior distributions of parameters had weird spikes indicating that the sampler got stuck somewhere.
* Some of the times the sampler got stuck, the two variance hyperparameters (`sigma` and `tau`) were at or near zero, which is not realistic.

I made the following changes to the model:
* I widened the prior on `mu` (std=4).
* I changed the priors on `sigma` and `tau` from `Exp(1)` to `Gamma(2, 1)`, which avoids zero and has a mode at 1.

The sampler still gets stuck occasionally when `mu` gets very low (-10), but this only happens when we have very few (zero?) viral read counts, in which case our posteriors aren't super useful except to give an upper bound on the P2RA coefficient.

I also refactored the `Model` class to simplify plotting and make it easier to adjust the hyperparameters.


jeffkaufman (approved):
Only a cursory review, but this seems reasonable

Issue #168, opened 2023-06-20
Summary figures


dp-rice:
Use the results of #167 to make some summary figures that compare viruses/mgs studies.

dp-rice:
Resolved by #174

PR #174, opened 2023-06-22
Make output summary table and figures

### LINKED ISSUE #167: Summary table
`fits_summary.txt` is not very readable. Make a TSV with high level summary information across pathogens. Specifically, summaries of the posterior distribution of the P2RA coefficient (on the biological scale).

dp-rice:
Resolves #167 and #168. Running `./fit.py` now creates two tables:

1. `fits.tsv` is a large table (not in the repo) with all of the posterior samples of the P2RA coefficients, for all locations and overall, on both the biological scale and the scale of the prior. This file is useful for plotting because it has all the information about the posterior.
2. `fits_summary.tsv` is a small table (in the repo) with descriptive statistics of the posterior distributions of P2RA coefficients (on the biological scale). This is useful for visual inspection.

There is also a new script `./plot_summaries.py` that uses `fits.tsv` to make two summary figures: one showing the distributions of the overall coefficients and the other showing the coefficients by location:
* [overall-boxen.pdf](https://github.com/naobservatory/p2ra/files/11839625/overall-boxen.pdf)
* [by_location-boxen.pdf](https://github.com/naobservatory/p2ra/files/11839624/by_location-boxen.pdf)


jeffkaufman (approved):
As noted verbally, it would be helpful if the ones that are downstream from "zero reads" were notated differently.  But that doesn't need to block submit.

PR #181, opened 2023-06-26
Simon mcv

simonleandergrimm:
Added MCV Virus. Note that there is also this [study](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2655709/#:~:text=We%20found%20differential,overall%20seroprevalence%20data.) that I looked into, but ended up not using for MCV. Let me know if you want to hear my reasoning why; I can jot it down (or explain in person).


jeffkaufman (approved):

pathogens/mcv.py
```diff
@@ -2,27 +2,48 @@
...
     taxid=TaxID(493803),
 )
 
+seroprevalence_women_us_2009 = Prevalence(
+    infections_per_100k=59.4 * 100_000,
+    number_of_participants=451,
+    country="United States",
+    date="2009",
+    active=Active.LATENT,
+    source="https://academic.oup.com/jnci/article/101/21/1510/963538#:~:text=A%20cut%20point%20of%2015%E2%80%89000%20median%20fluorescent%20intensity%20units%20was%20chosen%20as%20described%20above%20and%20used%20to%20identify%20268%20(59.4%25)%20of%20the%20451%20control%20subjects%20as%20seropositive%20for%20MCPyV%20VP1%3B%20this%20percentage%20was%20similar%20to%20that%20among%20women%20in%20control%20group%201%20(ie%2C%2053%25)."
```
jeffkaufman, line 19: This source also has info on some other viruses in the same 413 women; ex BKV at 92% and JCV at 45%.  We have BKV at 82% and JCV at 39% on more representative populations, so possibly we should push MCV down slightly?  But these are all very close, especially in terms of orders of magnitude!

PR #183, opened 2023-06-27
Extrapolated HPV prevalence to 2020 and 2021.

simonleandergrimm:


pathogens/hpv.py
```diff
@@ -49,12 +51,21 @@
...
+    return [
+        dataclasses.replace(
+            nhanes_2013_2016_18_59_yo_prevalence,
+            date_source=Variable(date="2020"),
+        ),
+        dataclasses.replace(
+            nhanes_2013_2016_18_59_yo_prevalence,
+            date_source=Variable(date="2021"),
+        ),
+    ]
```
jeffkaufman, line 68: I think while we're at it we should remove incidences; @dp-rice does that sound right?

simonleandergrimm: This got resolved in our conversation, as #184 already does this.

jeffkaufman (approved):

PR #184, opened 2023-06-27
Stopped returning incidences for HPV, HCV, and HBV. 

simonleandergrimm:
Stopped returning incidences for HPV, HCV, and HBV. Instead we should use prevalences. Removed HCV and HBV from `skip = []` in` pathogens/__init__.py`


jeffkaufman (changes requested):
Instead of commenting these out, just delete them.  We can restore them from source control if we want to later.

pathogens/__init__.py
```diff
@@ -22,7 +22,7 @@
 # For Hepatitis B and C, we're waiting on extrapolating older estimates
 # to the study period:
 # https://github.com/naobservatory/p2ra/pull/154
-skip = ["hbv", "hcv"]
+skip: List[str] = []
```
jeffkaufman, line 25: Just delete `skip`, here and below on lines 34-35

pathogens/hcv.py
```diff
@@ -155,27 +155,9 @@ def estimate_incidences():
     )
 
     estimates = [
-        acute_2019,
-        acute_2020,
-        acute_2021,
-        reported_acute_ohio_2020 * (acute_underreporting_factor),
-        reported_acute_ohio_2021 * (acute_underreporting_factor),
+        # For now we are dropping incidences.
```
jeffkaufman, line 158: Just `return []`

jeffkaufman (approved):

PR #185, opened 2023-06-27
update table to include nucleic acid, predictor type, and selection round

jeffkaufman:
image


simonleandergrimm (approved):

simonleandergrimm:
@jeffkaufman You likely need to briefly update this to incorporate AAV 5 and 6. Also HPV is currently turning up as an incidence, which I assume will be fixed automatically when rerunning the script.

PR #187, opened 2023-06-28
Simon aav update

simonleandergrimm:


pathogens/aav6.py
```diff
@@ -0,0 +1,63 @@
...
+
+pathogen_chars = PathogenChars(
+    na_type=NAType.DNA,
+    enveloped=Enveloped.NON_ENVELOPED,
+    taxid=TaxID(68558),
+)
+
+
+seroprevalence_hemophilia_global_2021 = Prevalence(
+    infections_per_100k=0.487 * 100_000,
```
jeffkaufman, line 19: I'm fine with this going in as a first cut, but since we're working on US data it seems more better to take the US-specific numbers from the paper.

pathogens/aav5.py
```diff
@@ -0,0 +1,59 @@
...
+
+seroprevalence_hemophilia_global_2021 = Prevalence(
+    infections_per_100k=0.348 * 100_000,
+    # Taking seropositivity from Figure 1A (un-weighted global seroprevalence)
+    number_of_participants=513,
+    # Though these participants are not representative of the general
+    # population, hemophilia is not caused by AAV-5. Prevalence would thus be
+    # at most affected by, e.g., lower socioeconomic status due to a higher
+    # disease burden.
+    country="Global",
```
jeffkaufman, line 25: I'm fine with this going in as a first cut, but since we're working on US data it seems more better to take the US-specific numbers from the paper.

PR #190, opened 2023-06-28
Improve summary figures

### LINKED ISSUE #119: Add more tests for stats.py
Including:

* `Model` should raise appropriate exceptions if you try to do stuff before model is fit
* `Model.fit_model` should initialize the `fit` and `dataframe` attributes

### LINKED ISSUE #176: Show viral read counts on boxen plots
Our posterior summary plots ([overall-boxen.pdf](https://github.com/naobservatory/p2ra/files/11840046/overall-boxen.pdf)) have a major limitation. Some of the viruses have very few (or zero) reads in some studies. In those cases, the left tail of the posterior is determined entirely by the prior, which is largely arbitrary and chosen for sampling purposes.

It would be nice if we could quickly see the number of viral reads on the same plot, so that we can mentally adjust for this.

### LINKED ISSUE #179: Display read counts on output figures
Right now the parameter estimates are somewhat hard to compare without without knowing which estimates are based on a lot of observations and which are driven mostly by the prior + zero or a few observed reads. We'd like to show this clearly on the figures.

dp-rice:
Improve summary figures in various ways to get the versions in the 2023-06-28 All Hands memo. (Resolves #176 and resolves #179, among other things.)

Also, adds some tests for model fitting now that that interface is more stable (Resolves #119).


dp-rice:
@jeffkaufman removing you as a reviewer and converting back to draft because (1) I realized that I broke your plotting script and (2) you're going away

plot_summaries.py
```diff
@@ -5,56 +5,199 @@
...
+        *ax.get_xlim(),
+        linestyle="dashed",
+        color="0.5",
+        linewidth=0.5,
+    )
+
+
+def italicize_incidence_viruses(ax) -> None:
+    # TODO: Get from data
+    incidence_viruses = ["SARS-COV-2", "Norovirus (GI)", "Norovirus (GII)"]
```
jeffkaufman, line 40: This is missing Flu A and B, but better to read from the data than fix here

jeffkaufman (approved):

PR #192, opened 2023-06-30
Extrapolated Northern European AAV seroprevalence in hemophilia patients to Denmark 2015-2018

simonleandergrimm:
Took the weigthed average of seroprevalence measurements in Northern European study participants, treating it as a Denmark estimate. Then extrapolated 2022 estimate to 2015-2018.


pathogens/aav2.py
```diff
@@ -36,6 +36,41 @@
...
 
 
+def denmark_extrapolated_seroprevalence() -> Prevalence:
+    # Taking weighted average of Northern European AAV-2 seropositivity
+    # numbers from Figure 1C, combined with participant numbers taken from the supplement.
+    seroprevalence_by_country = {
+        "France": {87: 0.605},
+        "Germany": {90: 0.483},
+        "United Kingdom": {17: 0.647},
+    }
```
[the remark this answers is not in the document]
jeffkaufman: You're just keying two pieces of data by country, so the extra layer of hashing isn't giving you anything except an implication that there might be multiple estimates for a country and more awkward retrieval API

pathogens/aav2.py
```diff
@@ -36,6 +36,41 @@
...
+    pairs: list[tuple[Prevalence, Population]] = []
+    for country, vals in seroprevalence_by_country.items():
+        for n_participants, seroprevalence in vals.items():
+            pairs.append(
```
jeffkaufman, line 50: Instead of:

```
for country, vals in seroprevalence_by_country.items():
        for n_participants, seroprevalence in vals.items():
```

you can now do:

```
for country, (n_participants, seroprevalence) in ...
```

pathogens/aav2.py
```diff
@@ -36,6 +36,41 @@
...
 
+def denmark_extrapolated_seroprevalence() -> Prevalence:
+    # Taking weighted average of Northern European AAV-2 seropositivity
+    # numbers from Figure 1C, combined with participant numbers taken from the supplement.
+    seroprevalence_by_country = {
+        "France": {87: 0.605},
+        "Germany": {90: 0.483},
+        "United Kingdom": {17: 0.647},
+    }
+    pairs: list[tuple[Prevalence, Population]] = []
```
jeffkaufman, line 47: `pairs` -> `prevalence_population_pairs`

pathogens/aav2.py
```diff
@@ -36,6 +36,41 @@
...
+                        source="https://www.ncbi.nlm.nih.gov/pmc/articles/PMC9063149/#:~:text=Seropositivity%20for%20(A)%20the%20global%20population",
+                    ),
+                    Population(
+                        people=n_participants,
+                        date="2022",
+                        country=country,
+                        source="https://www.ncbi.nlm.nih.gov/pmc/articles/PMC9063149/#:~:text=(93K%2C%20docx)-,Supplemental%20data%3A,-Click%20here%20to",
+                    ),
+                )
+            )
```
jeffkaufman, line 67: Probably simpler to use a list comprehension:

```
prevalence_population_pairs = [
    (Prevalence(...), Population(...))
    for country, (n_participants, seroprevalence) in ...
]
```

jeffkaufman: In general, anytime you're turning one collection into another collection with the same number of elements, and don't have any logic or complicated code inside, a list comprehension is a good bet

pathogens/aav2.py
```diff
@@ -36,6 +36,41 @@
...
+    return dataclasses.replace(
```
jeffkaufman, line 68: I think this would be clearer as `northern_european_average_seroprevalence()`, with the extrapolation to denmark in `estimate_prevalences`.  And if the extrapolation became complex, I would want to put that into a third method.

pathogens/aav2.py
```diff
@@ -54,9 +89,32 @@ def estimate_prevalences() -> list[Prevalence]:
         us_2020,
         date_source=Variable(date="2021"),
     )
+
+    # Extrapolating Denmark estimate backward in time to 2015-2018:
+    dk_2015 = dataclasses.replace(
+        denmark_extrapolated_seroprevalence(),
```
jeffkaufman, line 95: I would be a bit happier with:

```
dk_2022 = denmark_extrapolated_seroprevalence()
dk_2015 = dataclasses.replace(dk_2022, ...)
...
```

This has two advantages:

1. It makes it clearer how you are changing the date, 2022 to what?
2. It doesn't require running denmark_extrapolated_seroprevalence many times (which doesn't matter this case, because the method is very fast, but it still looks slightly weird to me to see an idempotent method called so many times)

pathogens/aav2.py
```diff
@@ -36,6 +36,41 @@
 )
 
 
+def denmark_extrapolated_seroprevalence() -> Prevalence:
+    # Taking weighted average of Northern European AAV-2 seropositivity
```
jeffkaufman, line 40: This should say what you are weighting your average by.

pathogens/aav2.py
```diff
@@ -36,6 +36,41 @@
...
+                    Prevalence(
+                        infections_per_100k=seroprevalence * 100_000,
+                        number_of_participants=n_participants,
+                        country=country,
+                        date="2022",
+                        active=Active.LATENT,
+                        source="https://www.ncbi.nlm.nih.gov/pmc/articles/PMC9063149/#:~:text=Seropositivity%20for%20(A)%20the%20global%20population",
+                    ),
+                    Population(
+                        people=n_participants,
```
jeffkaufman, line 61: Why are you weighting by number of participants?

jeffkaufman (changes requested):

pathogens/aav2.py
simonleandergrimm: I assume that the difference in seroprevalences between countries is largely driven by the small sample sizes, not by in-between-country differences. 

I'm accounting for this noise by weighting seroprevalence measurements by their sample size.

pathogens/aav2.py
simonleandergrimm: I instead did the following:
```
    # Extrapolating Northern European 2022 estimate to Denmark, backward in
    # time to 2015-2018:
    northern_europe_2022 = northern_european_average_seroprevalence()
    dk_2015 = dataclasses.replace(
        northern_europe_2022,
        date_source=Variable(date="2015"),
        location_source=Variable(country="Denmark"),
    )
   ...
```

jeffkaufman:
I see several points marked as "resolved" but without the underlying change made; do you need to push?

pathogens/aav2.py
```diff
@@ -36,6 +36,41 @@
 )
 
 
+def denmark_extrapolated_seroprevalence() -> Prevalence:
+    # Taking weighted average of Northern European AAV-2 seropositivity
+    # numbers from Figure 1C, combined with participant numbers taken from the supplement.
+    seroprevalence_by_country = {
```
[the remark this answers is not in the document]
jeffkaufman: looks not done?

pathogens/aav5.py
```diff
@@ -33,6 +33,45 @@
 )
 
 
+def northern_european_average_seroprevalence() -> Prevalence:
+    # Taking weighted average of Northern European AAV-5 seropositivity
+    # numbers from Figure 1D, combined with participant numbers taken from the
+    # supplement. We do this, because differences in seroprevalence between
+    # countries are likely driven by small sample sizes, not by in-between-
+    # country differences.
+
```
jeffkaufman, line 42: That seems unlikely to me?  5.9% with 17 people is 1/17, and if the UK were actually at a level near France or Germany you'd expect 5/17 or 6/17.  @dp-rice ?

pathogens/aav6.py
```diff
@@ -37,6 +37,44 @@
 )
 
 
+def northern_european_average_seroprevalence() -> Prevalence:
+    # Taking weighted average of Northern European AAV-5 seropositivity
+    # numbers from Figure 1D, combined with participant numbers taken from the
+    # supplement. We do this, because differences in seroprevalence between
+    # countries are likely driven by small sample sizes, not by in-between-
+    # country differences.
```
jeffkaufman, line 45: less worried about this one, since they're closer

pathogens/aav2.py
```diff
@@ -36,6 +36,44 @@
 )
 
 
+def northern_european_average_seroprevalence() -> Prevalence:
+    # Taking weighted average of Northern European AAV-5 seropositivity
+    # numbers from Figure 1D, combined with participant numbers taken from the
+    # supplement. We do this, because differences in seroprevalence between
+    # countries are likely driven by small sample sizes, not by in-between-
+    # country differences.
```
jeffkaufman, line 44: this one also seems fine

pathogens/aav5.py
simonleandergrimm: Yeah, you are right about this potentially being wrong. I could change this to a more conservative statement, like so:

"We aim to estimate AAV prevalence in a population that comes as close as possible to the Danish population. To do so, we're aggregating seroprevalence measurements from Northern European countries (France, UK, Germany), weighted by their participant counts. These measurements are from the same study as the one used above."

pathogens/aav2.py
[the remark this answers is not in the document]
simonleandergrimm: Missed this, sorry!

pathogens/aav5.py
simonleandergrimm: @jeffkaufman Let me know what you think. I can ping Dan again if his thoughts would be most useful.

jeffkaufman: I think it's not a big deal, and there's no reason to expect Denmark to be especially close to the UK

jeffkaufman (approved):

Issue #198, opened 2023-07-03
Revisiting DK EBV and CMV estimates, or adding limitation to report:


simonleandergrimm:
When extrapolating German or Dutch CMV/EBV estimates for Denmark, we need to think about the share of foreign-born in the extrapolated estimate and in the target location. E.g., foreign-born individuals have markedly higher CMV seroprevalence; if Copenhagen has large non-Western populations, seroprevalence there could thus be higher than in the general population of Germany or the Netherlands.

See this graph for the difference  in CMV seroprevalence between Dutch Western and non-Western individuals([study link, page 3](https://doi.org/10.1016/j.jcv.2014.11.033))
<img width="392" alt="image" src="https://github.com/naobservatory/p2ra/assets/58591538/3ef9b73b-d935-4fee-a84e-f40220d43ec7">

simonleandergrimm:
Will discuss with Charlie Whittaker how relevant this is.

PR #199, opened 2023-07-03
Extrapolated 2006 Dutch CMV seroprevalence to Denmark 2015-2018.

simonleandergrimm:
See issue #198 for a potentially relevant limitation that we might mention in the limitations of the report, or adjust for ,sometime in the future.

> When extrapolating German or Dutch CMV/EBV estimates for Denmark, we need to think about the share of foreign-born in the extrapolated estimate and in the target location. E.g., foreign-born individuals have markedly higher CMV seroprevalence; if Copenhagen has large non-Western populations, seroprevalence there could thus be higher than in the general population of Germany or the Netherlands.

### DIFF
--- a/pathogens/cmv.py
+++ b/pathogens/cmv.py
@@ -38,12 +38,13 @@ dutch_seroprevalence_2006 = Prevalence(
     infections_per_100k=0.456 * 100_000,
     confidence_interval=(0.419 * 100_000, 0.493 * 100_000),  # 95% CI
     active=Active.LATENT,
-    number_of_participants=6249,
+    number_of_participants=6382,
     # "This study used information from the PIENTER2 project, a crosssectional
     # population-based serum bank established in 2006–2007.[...]
     # A total of 19,781 individuals, including 2558 non-Western migrants, were
     # invited to complete a questionnaire and donate a blood sample. Serum
-    # samples were obtained from 6386 individuals (32%)."
+    # samples were obtained from 6386 individuals (32%). Sufficient serum to
+    # test for CMV antibodies was available from 6382 participants.
     source="https://doi.org/10.1016/j.jcv.2014.11.033",  # Page 4, heading 4.2
     country="Netherlands",
     date="2006",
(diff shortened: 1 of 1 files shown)


pathogens/cmv.py
```diff
@@ -32,7 +32,26 @@
     source="https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6059406/#:~:text=Overall%20CMV%20seroprevalence,Germany%20in%20women",
 )
 
-adult_seroprevalence_raleigh_durham_US = Prevalence(
+dutch_seroprevalence_2006 = Prevalence(
```
jeffkaufman, line 35: Maybe `nl_` instead of `dutch_`?

pathogens/cmv.py
```diff
@@ -75,17 +94,55 @@
...
+    # Extrapolating 2006 Dutch data to Denmark. The prevalence of the
+    # Netherlands looks roughly similar to estimates for Belgium (0.52%),
+    # Germany (0.45%) or Sweden (0.71%) as found in this review of global
+    # CMV seroprevalence:https://doi.org/10.1002/rmv.2034, Figure 2. Specific
+    # numbers are found in the supplement of the same study(https://
+    # onlinelibrary.wiley.com/action/downloadSupplement?doi=10.1002%2Frmv.2034&
+    # file=rmv2034-sup-0001-Appendix_S1.docx.
+    # Note the limitation noted in Issue #198: https://github.com/
+    # naobservatory/p2ra/issues/198
+    dk_2015 = dataclasses.replace(
```
jeffkaufman, line 107: I think it would be clearer what extrapolations you were doing if you did either:

```
dk_2006 = ...
dk_2015 = dataclasses.replace(dk_2006, ...)
dk_2016 = dataclasses.replace(dk_2006, ...)
...
```

Or you made 2016-2018 look like you currently have for 2015.

Otherwise it looks like 2016-2018 have less date extrapolation than they actually do

jeffkaufman (approved):

PR #200, opened 2023-07-03
HPV Denmark estimate, based on a 2016 study of 4000 Danish men. 

simonleandergrimm:
@jeffkaufman I will extend this estimate to cover all HPV subtypes covered in the study.

### DIFF
--- a/pathogens/hpv.py
+++ b/pathogens/hpv.py
@@ -32,7 +32,7 @@ dk_male_conscripts_seroprevalence_2016 = Prevalence(
     # The study participants were 18 to 65 years of age; 69.0% were younger
     # than 21 years (mean, 23 years) and 14.5% were 30 years or older."
     country="Denmark",
-    date="2015",
+    date="2016",
     active=Active.LATENT,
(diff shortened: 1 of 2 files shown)


jeffkaufman:
> I will extend this estimate to cover all HPV subtypes covered in the study

I can't tell whether this means you want me to hold off on reviewing this, or whether instead you're describing future work you're planning to do?

simonleandergrimm:
@jeffkaufman The issue is future work and shouldn't block this pull request.

pathogens/hpv.py
```diff
@@ -20,6 +20,22 @@
     selection=SelectionRound.ROUND_1,
 )
 
+dk_male_conscripts_seroprevalence_2016 = Prevalence(
+    infections_per_100k=0.418 * 100_000,
+    confidence_interval=(0.399 * 100_000, 0.438 * 100_000),  # 95% CI
+    # "The overall HPV prevalence was [...] 41.8% (95% CI, 39.9–43.8) with
+    # PCR. "
```
jeffkaufman, line 27: It looks like they did both HC2 and PCR; could you add something about why you think we should go with the PCR measurement?

jeffkaufman (changes requested):

pathogens/hpv.py
simonleandergrimm: Good q. Overall, PCR is a lot more sensitive:
"Most (97.0%) of the 1417 PCR-negative men were also HC2 negative, whereas 50.9% of the 1019 PCR-positive men were HC2 negative."

While digging into the study, I saw that PCR sometimes also catches unclassified HPV, called HPVX. But if such unclassified HPV turns up in wastewater and is sequenced, Kraken will likely classify it either as [unclassified Papillomaviridae](https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?lvl=0&id=333774), or will assign it to the family or genus level.

image

simonleandergrimm: Added a comment in the file.

jeffkaufman (approved):

PR #203, opened 2023-07-05
Update infra to cover Brinch 2020

jeffkaufman:
* rna_bioprojects is now target_bioprojects
* Brinch is two bioprojects, so in a few places we now need to handle there being multiple ones for a study
* Brinch doesn't get have public health estimates for all viruses, so I had to turn that off as well

### DIFF
--- a/stats.py
+++ b/stats.py
@@ -378,22 +378,28 @@ def choose_predictor(predictors: list[Predictor]) -> Predictor:
 
 def build_model(
     mgs_data: MGSData,
-    bioproject: BioProject,
+    bioprojects: list[BioProject],
     predictors: list[Predictor],
     taxids: frozenset[TaxID],
     random_seed: int,
+    enrichment: Optional[Enrichment],
 ) -> Model:
-    samples = mgs_data.sample_attributes(
-        bioproject, enrichment=Enrichment.VIRAL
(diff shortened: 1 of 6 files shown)


stats.py
```diff
@@ -378,22 +378,28 @@ def choose_predictor(predictors: list[Predictor]) -> Predictor:
...
     random_seed: int,
+    enrichment: Optional[Enrichment],
 ) -> Model:
-    samples = mgs_data.sample_attributes(
-        bioproject, enrichment=Enrichment.VIRAL
-    )
+    all_sample_attributes = {}  # sample -> attributes
+    study_viral_reads = {}  # sample -> viral_reads
+    for bioproject in bioprojects:
+        sample_attributes = mgs_data.sample_attributes(
```
dp-rice, line 390: Optional: rather than defining this intermediate variable, I'd rather put this directly in the argument to `update` so that we can rename `all_sample_attributes` -> `sample_attributes`. That way the naming is consistent and we can't accidentally use the wrong variable.

dp-rice (approved):
One optional suggestion. Otherwise, looks good.

stats.py
jeffkaufman: done!

PR #207, opened 2023-07-11
Fit brinch

dp-rice:
Fit the model to the Brinch data (as well as any changes that have been made to the pathogen files since the June all-hands memo).

Model fit:
overall-boxen

Matching reads table:
matching-reads-table

Note that this partially breaks `plot_summaries.py` but this will be revised significantly soon to improve the figures.


fit.py
```diff
@@ -22,14 +29,15 @@ def start(num_samples: int, plot: bool) -> None:
...
-        _,
         pathogen_name,
+        tidy_name,
         predictor_type,
         taxids,
         predictors,
     ) in predictors_by_taxid():
         taxids_str = "_".join(str(t) for t in taxids)
         for study, bioprojects in target_bioprojects.items():
+            print(pathogen_name, study)
```
jeffkaufman, line 40: is this intentional?

stats.py
```diff
@@ -110,7 +110,16 @@ class DataPoint(Generic[P]):
     sample: Sample
     attrs: SampleAttributes
     viral_reads: int
-    predictor: P
+    predictor: P | None
+
+    def get_predictor_value(self) -> float:
+        # TODO: Handle missing data
```
jeffkaufman, line 116: what's this TODO about?

jeffkaufman (approved):

fit.py
dp-rice: I was using it for debugging. Doesn't hurt anything, but it doesn't really help either. I'll remove it.

stats.py
dp-rice: If we eventually update the Stan model to allow for some of the samples to be missing the predictor data, we'll change this to return the value Stan is expecting for missing data (e.g., NaN). Comment updated

PR #214, opened 2023-07-14
Drop HSV-2, HBV, and AAV6

jeffkaufman:
See https://docs.google.com/document/d/1IIeOFKNqAwf9NTJeVFRSl_Q9asvu9_TGc_HSrlXg8PI/edit and https://github.com/BenLangmead/aws-indexes/issues/18

image

### DIFF
--- a/pathogens/aav6.py
+++ b/pathogens/aav6.py
@@ -117,14 +117,7 @@ def estimate_prevalences() -> list[Prevalence]:
         location_source=Variable(country="Denmark"),
     )
 
-    return [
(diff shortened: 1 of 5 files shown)


pathogens/aav6.py
dp-rice, line 1: It would be good to have a comment in the pathogen files explaining why they were dropped.

jeffkaufman: Added!

simonleandergrimm (approved):

PR #216, opened 2023-07-18
Plot estimates with zero counts differently

dp-rice:
For estimates with no viral reads, don't plot the boxes, instead indicate that we have upper bounds on our estimates by plotting:

* Left-pointing triangles at the ~94th percentile
* A faint dashed line to the left edge of the plot

incidence-boxen

prevalence-boxen

Note that for the plot by location, we will still plot the boxes since these estimates use information borrowed from the other locations.

by_location_incidence-boxen

One other minor change: we now give the vertical bar for the median the same alpha value as the other elements.

### DIFF
--- a/plot_summaries.py
+++ b/plot_summaries.py
@@ -80,7 +80,8 @@ def plot_boxen(
         hue="study",
         hue_order=plotting_order.study.unique(),
         showfliers=False,
-        box_kws={"linewidth": 0.5},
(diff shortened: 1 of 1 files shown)


jeffkaufman (approved):

dp-rice:
Final version:

incidence-violin

prevalence-violin

by_location_incidence-violin

PR #220, opened 2023-07-21
Run model with read exclusions

### LINKED ISSUE #218: Pathogens with subtypes only get model diagnostic plots for one
Because the two Influenzas and Noroviruses come from the same pathogen file and the pathogen file name is used to label the diagnostic output plots, the diagnostic plots from one subtype overwrite those from the other subtype.

dp-rice:
Bump the MGS data tag to remove false-positive read assignments from counts.

Updated prevalence plot:

prevalence-violin

Also, incidentally resolves #218


jeffkaufman:
Would it be possible to include Brinch for HIV and HCV, since those two reads look real?

dp-rice:
Recall that we don't have public health estimates for those. Otherwise they should be there automatically

dp-rice:
Also @jeffkaufman I forgot to push a commit

jeffkaufman:
Should this PR include `fits_summary.tsv`?

jeffkaufman:
> Recall that we don't have public health estimates for those. Otherwise they should be there automatically

pinging @simonleandergrimm ;)

dp-rice:
Ok, ready now.

jeffkaufman (approved):

PR #223, opened 2023-07-27
Fit Brinch HIV/HCV

dp-rice:
Updated prevalence figure:

prevalence-violin


dp-rice:
@jeffkaufman you can merge this if it looks good. Assigning to you in case you need to change anything as a result of the output.

jeffkaufman (approved):

PR #224, opened 2023-07-28
Switch from ra-1000 to ra-100

jeffkaufman:
We previously estimted RA(1:1000), but decided to switch to RA(1:100) to avoid the per-mille symbol.

I needed to change several places in the figures where there were items placed relative to the x-axis, but I think I got them all.

I also tweaked the x-axis labeling slightly, so now it says "at 1% prevalence" or "at 1% incidence".

I needed to set the dpi on export or else I got pngs that were tiny and blurry.  Possibly you've set a default DPI somewhere not in the repo?

image

image

image


dp-rice (approved):
Looks good! And the dpi probably should have been specified there to begin with.

PR #225, opened 2023-08-01
Pathogens: remove LA-specific HIV estimate

jeffkaufman:
Early in our process of collecting data we found an HIV estimate for LA County.  We don't have other county-specific estimates, including for nearby sampled counties like Orange County and San Diego County.  Drop the LA-specific estimate for consistency.

This only affects our estimates for HTP, SJ, and JWPCP in the Rothman data, and switches them to use our US-wide estimate of 157 per 100k from an LA-specific estimate of 231 per 100k.

You can just barely see the difference in the violin plot:

image


simonleandergrimm (approved):
Looks good!

PR #226, opened 2023-08-01
Add bar chart for prevalence

jeffkaufman:
image


simonleandergrimm (changes requested):
Looks good! Things I would change:
- I would drop the top and right spine of the plot (we do this for most plots in the report) 
- I would change the title to "Estimated prevalence of persistent viral infections"
- You could drop the x-axis label, or replace it with "Prevalence".
- I would change the label to "Country, Year (Study)" as the former is the main distinction between the blue and orange bar."
- Also, having a tiny bit of space between the within-pathogen bars could be worth trying

jeffkaufman:
@simonleandergrimm done!

image

simonleandergrimm (approved):

PR #231, opened 2023-08-07
Also generate by_location_prevalence-violin plot

jeffkaufman:


dp-rice (approved):
We did this together.

PR #235, opened 2023-09-06
Plot reads per prevalence

dp-rice:
Make a new figure (not in the P2RA report) to illustrate the cost estimates.

`./figures/plot_reads_per_prevalence.py` generates `figures/reads_per_prevalence.png`:
reads_per_prevalence


plot_reads_per_prevalence.py
```diff
@@ -0,0 +1,83 @@
...
+            )
+            ax.fill_between(
+                incidence,
+                scale_factor / upper,
+                scale_factor / lower,
+                color=color,
+                alpha=0.2,
+            )
+        ax.set_title(virus)
+        ax.set_xlabel("Prevalence")
```
jeffkaufman, line 72: Maybe supxlabel to share this xlabel between the three plots?

plot_reads_per_prevalence.py
```diff
@@ -0,0 +1,83 @@
...
+                max=float(row["max"]),
+            )
+    return data
+
+
+def start():
+    data = read_data()
+    viruses = [
+        ("Norovirus (GII)", "incidence"),
+        ("SARS-COV-2", "incidence"),
```
jeffkaufman, line 44: We'e mostly been doing "Sars-CoV-2"

plot_reads_per_prevalence.py
jeffkaufman, line 72: Shouldn't these be "Weekly Incidence" not "Prevalence", at least for the first two?

jeffkaufman (approved):

reads_per_prevalence.png
jeffkaufman, line 1: We don't usually check images into git

plot_reads_per_prevalence.py
dp-rice: They should but this was something Will requested for the original use of the figure (in a report that was eliding the difference)

dp-rice:
@jeffkaufman thanks! For context, this was a quick figure that Will asked for for his report that Simon is planning to adapt for the publication. I've asked him to review the pull request so he's familiar with the code and can ask questions about it.

simonleandergrimm (approved):
Looks good! I would just move the figure script into the figures directory.

There are some aesthetic details I would change (drop top and right axes, drop ticks on middle and right-most plot), but I can do this when I create a similar figure, as @dp-rice and I discussed yesterday.

PR #236, opened 2023-10-11
Changed logic of hatch_zero_counts.

simonleandergrimm:
Previously, when `hatch_zero_counts` was set to False, the code would add hatching. I reversed the logic to hatch when set to True.

PR #237, opened 2023-11-11
Added mypy inline comments and fixed tuple mypy issue in plot_summaries.py

simonleandergrimm:
@jeffkaufman I fixed the mypy issue for now by adding `# type: ignore statements` to the lines in `plot_summaries.py` that caused trouble. let me know if this is an adequate fix for now. Also pinging @dp-rice for awareness.


simonleandergrimm:
@jeffkaufman I merged the commit made in this branch into a different branch (see #236), which I then merged with main (as it fixed the outstanding mypy issue that Dan told me to resolve before merging). Let me know if that was a premature decision.

PR #238, opened 2023-11-27
Set pseudocounts back to 0.1 infections per week (previously 0.001). Reran fit.py.

simonleandergrimm:
I noticed that our pseudocount value was too low (0.001 infections per week), even though it should be 0.1 infections per week. I failed to spot this in #229, where, as part of a sensitivity analysis, 0.1 was changed to 0.001 and subsequently committed to main. I now fixed this.

I reran run.py, which created a new `fits_summary.tsv`. @jeffkaufman Once you approve these changes I can replace the violin figures in the [P2RA report](https://naobservatory.org/reports/predicting-virus-relative-abundance-in-wastewater/#fnref-12)


jeffkaufman:
Thanks for catching and fixing my mistake!

jeffkaufman (approved):

PR #239, opened 2024-03-15
flu: check in data from current season

jeffkaufman:


simonleandergrimm (approved):
Looks good, the only changes I see are formatting changes across a couple of pathogens and more data on influenza

PR #240, opened 2024-03-15
Johnson Flu A: hack together a rough P2RA estimate

jeffkaufman:
This should not actually be merged, but it would be helpful for you to look at it to check the calculation is correct.  If we do like this then later we can figure out how to get things in here that depend on NAO-internal sequencing results.

Mean RA_i(1%) is 1.7e-8.


pathogens/influenza.py
```diff
@@ -140,6 +140,19 @@ def load_weekly_data() -> WeeklyData:
     source="https://www.cdc.gov/flu/about/burden/2021-2022.htm#:~:text=The%20overall%20burden%20of%20influenza%20(flu)%20for%20the%202021%2D2022%20season%20was%20an%20estimated%C2%A09%20million%C2%A0flu%20illnesses",
 )
 
+infections_2023_2024 = IncidenceAbsolute(
+    annual_infections=41_500_000,
+    confidence_interval=(29_000_000, 54_000_000),
```
dp-rice, line 145: This is a technical quibble, but I can't tell from glancing at the website whether their estimated range is technically a confidence interval.

dp-rice (commented):
@jeffkaufman this seems like you're doing a pretty straightforward adaptation of what you were doing before, right? I don't see any obvious problems

pathogens/influenza.py
jeffkaufman: You're right, it may not be.

I've used the midpoint of the two numbers from the website; the range isn't currently used for anything.

jeffkaufman:
> a pretty straightforward adaptation of what you were doing before, right?

That's right, though there is a bit of messiness about the flu estimate being from only a partial season.

